PR #1, opened 2024-05-08
Add vitest and first tests to project

laurenrevere:
**Purpose**
Add vitest and react testing library to the project. Add the first tests so we can build off of them.

**Notes**
Before merge, I need to fix an ESLint problem with `vitest.config.ts.` Would love some help @nicholaslyon-dt if you have the time.

**Next Steps**
- Update the web project test to actually test the application.
- Decide general test structure.


apps/web/vitest.config.ts
nicholaslyon-dt, line 1: Since these config files are the same for this and the ui package, I think the paradigm here is to move these to a config package (like eslint and typescript).

package.json
```diff
@@ -5,14 +5,20 @@
     "build": "turbo build",
     "dev": "turbo dev",
     "lint": "turbo lint",
-    "format": "turbo format"
+    "format": "turbo format",
+    "test": "vitest"
```
nicholaslyon-dt: The turborepo paradigm here is to set this to `turbo test`, update the `turbo.json` file with:

```
"test": {
  "dependsOn": ["^test"]
},
```

 and then add a "test": "vitest" to each workspace that needs to run tests.

package.json
```diff
@@ -5,14 +5,20 @@
...
     "@repo/typescript-config": "*",
+    "@testing-library/jest-dom": "^6.4.5",
+    "@testing-library/react": "^15.0.7",
+    "@vitejs/plugin-react": "^4.2.1",
+    "jsdom": "^24.0.0",
     "lefthook": "^1.6.10",
     "prettier": "^3.2.5",
-    "turbo": "latest"
+    "turbo": "latest",
+    "vitest": "^1.6.0"
```
nicholaslyon-dt: Then, all these vitest and related npm packages would be installed in each worksapce that needs them. I _**think**_ the theory here is that each workspace should include the npm packages that itself needs independently.

nicholaslyon-dt (commented):
Added a few comments related to the structure. Let me know what you think!

apps/web/vitest.config.ts
laurenrevere: Will the relative paths for the setup files work correctly?

They will need different setup files in practice (even if they keep the same name)

nicholaslyon-dt: Yeah I was thinking we could keep the common elements in a base file, then extend/update that object for specifics for each. But if in the end they won't overlap much LMK. Happy to leave them separate if they won't really duplicate stuff in the end.

packages/ui/vitest.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineProject, mergeConfig } from "vitest/config";
+import configShared from "../../vitest.shared.ts";
+
+export default mergeConfig(
```
laurenrevere: @nicholaslyon-dt added a config at the root and then merged it in the individual folders. I have some concern about the reliance on the shared config here. I might prefer going back to one config per workspace. Thoughts?

My understanding is that I can't automatically inherit from the root config. https://vitest.dev/guide/workspace#configuration

nicholaslyon-dt: Yeah I wasn't really thinking of the shared part as a 'config' per se, just a JS object that we could import for the reusable parts. But if this is the way they suggest in vitest, I'm fine with going back to one per workspace. It's easier to understand and sometimes that's best!

nicholaslyon-dt:
Only 2 things, the second of which I could take or leave:
1. Looks like the package-lock.json at the root of the project needs updated since the packages moved around.
2. I'm seeing a deprecated warning in the console when running the tests. Thoughts on fixing if it's easy? Points to this link:
https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated

Screenshot 2024-05-09 at 4 44 44 PM

nicholaslyon-dt (approved):

PR #3, opened 2024-05-10
FE-1: Basic Components

nicholaslyon-dt:
**Purpose**
- Create basic reusable components for across the site
  - Button
  - Link
  - Icon

[Figma](https://www.figma.com/design/fABseEsz7QvzaUXz9lRyPw/BCBC---Design-Discovery?node-id=471-16502&t=rOCCmu7sDfTFYseE-4)

**Notes**
- If built locally, the root page has an example for most of the component cases used across the site.


packages/constants/README.md
jbook24, line 1: Is the readMe confirmed to be up to date with the required steps to run the application? I wonder if we should mention the required node version usage? 

image

packages/ui/src/variables.css
jbook24, line 1: Was there a consideration to use SCSS?

packages/ui/src/icon/icons/ArrowOutwardIcon.tsx
jbook24, line 1: Thoughts on this being named either leftArrowIcon or rightArrowIcon?

packages/ui/src/link/Link.test.tsx
jbook24, line 1: Is this name referencing a hyperlink icon?

apps/web/app/globals.css
jbook24, line 1: Similar comment as below, was there a consideration of leveraging SCSS?

packages/ui/src/icon/icons/OpenInNew.tsx
jbook24, line 1: Is this name referencing a hyperlink icon?

packages/constants/README.md
nicholaslyon-dt: It's in the `package.json` at the root level! I would tend to suggest leaving it there to not have to duplicate documentation. Thoughts?

Screenshot 2024-05-10 at 4 03 30 PM

packages/ui/src/variables.css
nicholaslyon-dt: Not really. Vanilla CSS is so much more powerful than it used to be so the use case for pre-processors is smaller IMO. On a larger project that might be doing fancier design things it might be worth it, but for our use case I didn't figure we needed it.

packages/ui/src/icon/icons/ArrowOutwardIcon.tsx
nicholaslyon-dt: This is what material design called the icon so I kept it as that. I would assume the left and right arrow icons would point directly left or right respectively.

packages/ui/src/icon/icons/OpenInNew.tsx
nicholaslyon-dt: Same note as above. This icon is from the material design library and I used their naming convention. It's used in cases where the target is _blank on the link so it opens in a new window.

jbook24 (approved):

packages/ui/src/icon/Icon.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+    expect(queryByTitle("arrow pointing outward")).not.toBeInTheDocument();
+  });
+  // openInNew
+  it("renders openInNew icon with title", () => {
+    const { getByTestId, getByTitle } = render(
+      <Icon type="openInNew" id="openInNewIcon" title="open in a new tab" />,
+    );
+    const openInNewIcon = getByTestId(getIconTestId("openInNew"));
+    expect(openInNewIcon).toBeInTheDocument();
+    expect(openInNewIcon).toHaveAttribute("aria-labelledby", "openInNewIcon");
```
laurenrevere, line 73: Are we checking the value of `aria-labelledby` specifically in the tests that include a title to prove that it doesn't match the `title`?

Would it make sense to confirm the `aria-labelledby` value exists and is correct on on non-titled icon as well?

packages/ui/src/link/Link.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+// workspace
+import Link, { LinkVariant } from "./Link";
+import { getIconTestId } from "../icon/Icon.test";
+
+// test constants
+const TEXT = "Press me";
+const getLinkTestId = (variant: LinkVariant | string) =>
+  `${TESTID_LINK}-${variant}`;
+
+describe("Button", () => {
```
laurenrevere: Should be "Link" here.

laurenrevere:
Looks good. 1 typo and one question. I expect I'll have more testing opinions as the project goes on.

I believe that there's a build issue on main. Would we want to solve it here before merge?

packages/ui/src/icon/Icon.test.tsx
nicholaslyon-dt: As currently set up, the `aria-labelledby` attribute is only added on ones that also have a title so we're only checking for it in the ones where both attributes are provided. My current thought process is that I'm adding them together to add a11y text for screen readers. However, in some cases the icon could be decorative and not need to be read out (I'm thinking buttons with text AND and icon). So in those cases I wouldn't want the icon to have an aria-labelledby attribute. Thoughts?

nicholaslyon-dt:
> Looks good. 1 typo and one question. I expect I'll have more testing opinions as the project goes on.
> 
> I believe that there's a build issue on main. Would we want to solve it here before merge?

100%! Just double commenting here as well as in Slack. I've fixed the issue and pushed up some changes. LMK if it's woring for you now too!

packages/ui/src/icon/Icon.test.tsx
laurenrevere: Makes sense!

Valid, yes. I think the only time we have to be sure that there is an `aria-labelledby` is in this title situation, when an icon adds extra information, or when it's an icon button. Solid.

laurenrevere:
Noticing that buttons with icons are being called button groups in voicover. I wonder if this could cause some confusion for screen reader users. Any thoughts?

nicholaslyon-dt:
> Noticing that buttons with icons are being called button groups in voicover. I wonder if this could cause some confusion for screen reader users. Any thoughts?

Updated the icons to hide them. The titles are still read if they are there.

laurenrevere:
Small UI Polishes (further testing to come later):
- The focus ring for the active state is not directly on the boarder of the button like the designs. Is that something we want to adjust?
- The highlight of links is off-center when they have an icon. (I suggest this as polish for later.)

nicholaslyon-dt:
> Small UI Polishes (further testing to come later):
> 
> * The focus ring for the active state is not directly on the boarder of the button like the designs. Is that something we want to adjust?
> * The highlight of links is off-center when they have an icon. (I suggest this as polish for later.)

1. Yeah I moved it out because I felt the blue focus color was really hard to see against the blue button for usability purposes
2. I'm not 100% sure what this is referring to. Let's chat about it.

laurenrevere:
**Build**: fcd7049f101e8913922c60dd29cbeea95c2fdabd (and an earlier build with additions to the page)
**Device Info**: Safari and Chrome on Sanoma 14.4.1, Edge on Windows 11, Safari on an iPhone 11 (OS 17.0.2), Samsung Internet on a Google Pixel 4 (Android 10)

Reviewed buttons, links, and icons for color contrast, touch target, screen reader, and design.



**Observations/Notes**:
- When I tried samsung internet and talkback on the Pixel, I was not able to hear the titles of the icon buttons. This may require further investigation.
- Disabled buttons don’t pass color contrast checks and aren’t read by voiceover. This passes WCAG because they are not actionable. I added an improvement ticket to use aria-disabled so that disabled buttons show up in voiceover.

Screenshot 2024-05-15 at 1 17 43 PM

PR #5, opened 2024-05-14
TE-2: add cypress

laurenrevere:
**Purpose**
Adds cypress to the project and the first cypress test. Also adds the ability to run axe accessibility tests.

~I also added all tests to the prepush hooks. @nicholaslyon-dt curious if you're for or against that move.~
Added vitests to pre-commit and cypress tests to post-commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,11 @@ As noted above, this repository uses [Lefthook](https://github.com/evilmartians/
 
 #### Pre-Commit
 
-Pre-commit the `format` and `lint` turbo tasks will run. This will run any `scripts` called `format` or `lint` in any of the workspaces. Look at the `package.json` in each workspace for what executes in each. If there are any errors, the commit will be aborted.
+Pre-commit the `format`, `lint`, and `test` turbo tasks will run. This will run any `scripts` called `format`, `lint`, or `test` in any of the workspaces. Look at the `package.json` in each workspace for what executes in each. If there are any errors, the commit will be aborted. (Note: `test` normally includes the vitests.)
+
+#### Post-Commit
+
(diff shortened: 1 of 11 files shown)


nicholaslyon-dt:
@laurenrevere I could go either way. I like the idea of doing it pre-push to catch these things earlier (vs waiting to open a PR, seeing if they pass, then going forward with said PR).

However, we'll have to pay attention to timing here. If they get too long it might be cumbersome to wait for every commit. I'm mostly thinking in terms of just submitting an update/fix to a PR. We'd have to wait for them to run before we push up, then wait for them to run in the PR before we can merge.

I definitely think the unit tests should be run. I'm more thinking on timing of Cypress tests. And also this then means we have to have our local build running any time we push correct?

Thoughts?

laurenrevere:
> @laurenrevere I could go either way. I like the idea of doing it pre-push to catch these things earlier (vs waiting to open a PR, seeing if they pass, then going forward with said PR).
> 
> However, we'll have to pay attention to timing here. If they get too long it might be cumbersome to wait for every commit. I'm mostly thinking in terms of just submitting an update/fix to a PR. We'd have to wait for them to run before we push up, then wait for them to run in the PR before we can merge.
> 
> I definitely think the unit tests should be run. I'm more thinking on timing of Cypress tests. And also this then means we have to have our local build running any time we push correct?
> 
> Thoughts?

I agree with you that timing is a concern. And yes, the server would have to be running in order for the tests to pass. I was thinking that we could keep them running as a commit hook either until they got annoying or until we had CI set up. If we don't have a good space to run the tests then they're really not as useful imo.

I was also thinking we could make the cypress tests `post-commit`, but that makes them almost as annoying as pre-commit because it still takes over your command line.

nicholaslyon-dt (approved):

PR #6, opened 2024-05-15
Dockerize Application

clay-comer-wt:
Setup a Dockerfile and docker-compose.yml file for the application based on the example from [Turborepo](https://github.com/vercel/next.js/blob/canary/examples/with-docker/Dockerfile). 

**Docker Desktop**
Screenshot 2024-05-15 at 4 08 25 PM

**Local Browser**
Screenshot 2024-05-15 at 4 08 37 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ docker-compose build
 docker-compose up
 ```
 
+**NOTE**: Some security applications such as ZScaler may require additional configuration of the Docker install to 
+utilize a security certificate file. Please read [this documentation](https://docs.docker.com/engine/security/certificates/) 
+from the Docker website to learn more.
+
 ### Hooks
 
 As noted above, this repository uses [Lefthook](https://github.com/evilmartians/lefthook) to run git hooks.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -49,6 +50,18 @@ To build all apps and packages, run the following command:
 npm run build
 ```
 
+### Running in Docker
+
+To run the application locally in a Docker container, run the following commands:
+
+```
+docker-compose build
```
laurenrevere, line 63: Should we mention somewhere in the readme that users need to install docker?

Or list it as a prereq?

clay-comer-wt: That's a good point. Its an implied prereq, however we should be explicit

clay-comer-wt: @laurenrevere updated!

nicholaslyon-dt: Should we also note about the zscaler issue? Otherwise looks good, got it to build now!

nicholaslyon-dt (approved):

README.md
clay-comer-wt: I'll add a quick note about it

clay-comer-wt: @nicholaslyon-dt updated!

PR #7, opened 2024-05-15
FE-3: Radio Group Component

nicholaslyon-dt:
**Purpose**
Add a reusable Radio Group component for use with the walkthrough questions

**Notes**
1. I refactored the data a little bit after using it in the component. All `values` are strings now.
2. I tweaked some icons to make the more consistent
3. I added a new button variant for the `glossary` style since those will also be buttons that open the side model instead of links

**Testing**
I've added an example radio button component to the root page with all the other examples. It uses data being pulled in from the 9.9.9.json walkthrough file. It is unselected by default, but required, so if submit is pressed before selection it will show the error message.

### DIFF
--- a/packages/ui/src/button/Button.css
+++ b/packages/ui/src/button/Button.css
@@ -127,3 +127,24 @@
   width: 1em;
   transition: var(--transition-link-icon-fill-color);
 }
+
+/* Variant - Glossary */
+.ui-Button.glossary {
+  background-color: var(--surface-color-tertiary-button-default);
+  color: var(--typography-color-link);
+  text-underline-offset: var(--link-underline-offset);
+  text-decoration: dotted underline;
+  min-height: auto;
+  padding: 0;
+  border-radius: 0;
+
+  &[data-hovered] {
+    color: var(--theme-blue-80);
+    background-color: var(--theme-blue-20);
+  }
+
+  &[data-pressed] {
+    color: var(--typography-color-primary);
+    background-color: var(--theme-blue-40);
+  }
+}
--- a/packages/ui/src/radio-group/RadioGroup.test.tsx
+++ b/packages/ui/src/radio-group/RadioGroup.test.tsx
@@ -27,6 +27,7 @@ const getRadioGroupErrorTestId = (testId = RADIO_GROUP_NAME) =>
 const RADIO_GROUP_LABEL = "Number of storeys";
 const OPTION_1_VALUE = "1";
 const OPTION_2_VALUE = "2";
+const OPTION_3_VALUE = "3";
 const RADIO_GROUP_OPTIONS = [
   { label: "1 storey", value: OPTION_1_VALUE },
   { label: "2 storeys", value: OPTION_2_VALUE },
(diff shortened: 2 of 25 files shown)


packages/ui/src/button/Button.css
```diff
@@ -127,3 +127,24 @@
   width: 1em;
   transition: var(--transition-link-icon-fill-color);
 }
+
+/* Variant - Glossary */
```
laurenrevere, line 131: I really appreciate you paying attention to buttons vs links! 💯

packages/ui/src/radio-group/RadioGroup.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+        label={RADIO_GROUP_LABEL}
+        name={RADIO_GROUP_NAME}
+        defaultValue={defaultValue}
+      />,
+    );
+
+    // get option 2
+    const option2 = getByTestId(getRadioTestId(RADIO_GROUP_NAME, defaultValue));
+    const input2 = option2.querySelector("input[type='radio']");
+    expect(input2).toBeChecked();
```
laurenrevere, line 97: Would we also want to check that option 1 is not checked?

packages/ui/src/radio-group/RadioGroup.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+        value={selectedValue}
+      />,
+    );
+
+    // get option 3
+    const option3 = getByTestId(
+      getRadioTestId(RADIO_GROUP_NAME, selectedValue),
+    );
+    const input3 = option3.querySelector("input[type='radio']");
+    expect(input3).toBeChecked();
```
laurenrevere, line 127: We might want to confirm that the other two are not selected here as well.

laurenrevere (approved):
Small possible change on the tests, but other than that looks good!

nicholaslyon-dt:
@laurenrevere Done and done!

laurenrevere:
**Build**: 
Device Info: Safari on Sanoma 14.4.1, Edge and Chrome on Windows 11, Safari on an iPhone 11 (OS 17.0.2), Samsung Internet on a Google Pixel 4 (Android 10)

Reviewed radio button for:
- Design ❔
- Voiceover ✅
- Color Contrast ✅
- Keyboard Navigation ✅

**Observations/Notes**:
- Do we have an example of how to add a background to a radio button as shown[ in the designs](https://www.figma.com/design/fABseEsz7QvzaUXz9lRyPw/BCBC---Design-Discovery?node-id=471-16526&t=eHVcCZn6eYb5kgoN-4) @nicholaslyon-dt?
- When you navigate with the arrow keys, the selected value is changed. This is expected behaviour. (When you tab away from the input the last selected item stays selected.)
- Voiceover on mobile safari calls radio buttons both checkboxes and radio buttons. This appears to be a mobile safari quirk as it happens on radio buttons from accessibility examples such as: this example from [MagentaA11y](https://www.magentaa11y.com/checklist-web/radio/).

NVDA Example:
Screenshot 2024-05-17 at 2 25 39 PM

Voiceover Example:
Screenshot 2024-05-17 at 4 26 58 PM

nicholaslyon-dt:
@laurenrevere I'm not 100% sure what you mean by adding a background like in the designs. Do you mean the hover state?

Screenshot of 3 states
Screenshot 2024-05-20 at 9 51 34 AM

laurenrevere:
Yes, that is what I meant. I was focusing so much on keyboard navigation that I missed it, Thanks.

PR #8, opened 2024-05-17
Add initial docker build workflow step on PR and main merge

clay-comer-wt:
Add a starter GitHub workflow that does the following:

1. Checkouts out the repository from the current branch/pr
2. Builds the Docker container of the web application for the house-innovations-housnav project and tags it with a unique commit hash.
3. On merge to main, tag build as latest. (We can always change this depending on how we want to deploy, this is just a starting place.)

Testing:
Build details: https://github.com/bcgov/House-Innovations-HOUSNAV/actions/runs/9133784415/job/25117995946?pr=8

### DIFF
--- a/.github/workflows/docker-build-and-tag.yml
+++ b/.github/workflows/docker-build-and-tag.yml
@@ -27,7 +27,8 @@ jobs:
       - name: Build Docker image with commit hash
         run: docker build -t house-innovations-housnav-web:${{ env.GIT_SHA }} -f apps/web/Dockerfile . --no-cache
 
-      - name: Tag Docker image as latest
+      - name: Tag Docker image as latest for main branch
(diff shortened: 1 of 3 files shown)


nicholaslyon-dt (commented):
Looks great! I would think we can tweak if we need to as we go. Where does the tagged image ... go? live? Is this something that will work with Artifactory as we go forward?

clay-comer-wt:
> Looks great! I would think we can tweak if we need to as we go. Where does the tagged image ... go? live? Is this something that will work with Artifactory as we go forward?

That's correct. Tagged image will get pushed to Artifactory and stored there. So we tag the image with a unique commit hash, then also tag it as "latest" which is then pushed. We'll have to add some logic so that if we need to rollback we know what the previous "latest" was, but that shouldn't be too bad.

nicholaslyon-dt (approved):

PR #9, opened 2024-05-17
FE-4: Walkthrough Question Setup and Question 1

nicholaslyon-dt:
**Purpose**
Be able to navigate to the walkthrough and show question #1

**Notes**
- This includes a decent amount of walkthrough setup so any feedback on the structure of that would be beneficial as well.
- I also removed the `boolean` question type as it was the same as a Multiple Choice question with only 2 possible answers.
- Added a basic 404 page at the root for urls that aren't found. There is currently no design for this type of page.
- I added some of the walkthrough based components directly to the web application instead of in the ui since I felt they were specific to the application we're building and not something likely to be in a reusable component library.
- I left a lot of TODOs in here based on things that weren't in the scope of this ticket but we'll come back to as a part of future work.

**Testing**
1. Click link on main page to head to walkthrough 9.9.9
2. See question 1
3. Edit URL to go to another random walkthrough and see error page
4. Edit URL to go to /walkthrough and see redirect to homepage


clay-comer-wt (approved):
LGTM!

laurenrevere (approved):

laurenrevere:
See updated/completed comments on Jira: https://hous-bssb.atlassian.net/browse/HOUSNAV-5

PR #10, opened 2024-05-21
FE-5: Header Component

nicholaslyon-dt:
**Purpose**
- Create a reusable header component to display on all pages of the web application
- Includes mobile, tablet, and desktop displays

**Testing**
1. See header on home page
2. Click to walkthrough and see same header
3. Click "Home" in header to see home page
4. Click "Download" in header to download the full PDF of the building code
5. Switch to mobile device OR resize browser to see different screen size displays
6. Open mobile navigation and see same links as on desktop

**Notes**
The "Download" link currently points to the hosted PDF file on BC Govs website. Do we think this is sufficient?


apps/web/app/layout.tsx
```diff
@@ -3,14 +3,20 @@ import type { Metadata } from "next";
...
+  ID_MAIN_NAVIGATION,
+} from "@repo/constants/src/ids";
 import "@repo/ui/cssVariables";
 // local
 import { ClientProviders } from "./ClientProviders";
 import "./globals.css";
 
 export const metadata: Metadata = {
-  title: "BC Gov - Building Code - Walkthroughes",
+  title: "BC Gov - 2024 Building Code - Walkthroughes",
```
jbook24: Typo on Walkthroughs?

apps/web/app/page.tsx
```diff
@@ -3,12 +3,13 @@
...
 // repo
 import Link from "@repo/ui/link";
+import { URL_WALKTHROUGH_HREF } from "@repo/constants/src/urls";
 
 export default function Page(): JSX.Element {
   return (
     <div className="container">
       <p>
-        <Link href="/walkthrough/9.9.9/">Walkthrough 9.9.9</Link>
+        <Link href={`${URL_WALKTHROUGH_HREF}/9.9.9/`}>Walkthrough 9.9.9</Link>
```
jbook24, line 12: Do we think this string will be used in more than one place? should we create a constant here?

packages/ui/src/icon/icons/BCLogo.tsx
jbook24, line 1: Maybe not now, but should we explore the file size of the svg's used? are we following best practices?

apps/web/components/question/Question.css
```diff
@@ -17,6 +17,10 @@
   font: var(--typography-regular-large-body);
 }
 
+.web-QuestionTitle .ui-Button {
```
jbook24, line 20: for css naming conventions, do we want to standardize one of the following naming conventions? 
https://dev.to/ziizium/css-naming-conventions-5gd6

- BEM
- SMACSS
- ITCSS
- OOCSS
- AMCSS

packages/ui/src/header/Header.css
jbook24, line 1: for css naming conventions, do we want to standardize one of the following naming conventions?
https://dev.to/ziizium/css-naming-conventions-5gd6

BEM
SMACSS
ITCSS
OOCSS
AMCSS

apps/web/app/layout.tsx
nicholaslyon-dt: Weird. My IDE was saying Walkthroughs was spelled wrong and wanted the es. Added to the dictionary and fixed across the board.

apps/web/app/page.tsx
nicholaslyon-dt: Which part of the string are you referring to? Are you referring to the additional 9.9.9 part?

packages/ui/src/icon/icons/BCLogo.tsx
nicholaslyon-dt: To my knowledge yes. This is the only one larger than 1kb. But good call on this one. I did want to revisit it anyway because it looks blurry, which seems weird from an SVG perspective. I'm going to switch it to a PNG. Get's it down to 2kb, uses next/image, and looks better.

apps/web/app/page.tsx
jbook24: Yes, the 9.9.9 part? I'm okay either way

packages/ui/src/header/Header.css
nicholaslyon-dt: Yeah I can. I've never found them to be very helpful but in the past I've used a modified version of BEM with larger projects and larger teams. I'll update the current styles.

apps/web/app/page.tsx
nicholaslyon-dt: Possibly, but I also see this part being refactored when we actually build this landing page and the cards on it. I'd rather wait until later to see if we need to introduce a const until then. Because honestly the one that's there is only used once now too and probably not needed at this point.

jbook24 (approved):

packages/ui/src/header/Header.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      GET_TESTID_BUTTON(TESTID_HEADER_MOBILE_NAV_BUTTON),
+    );
+    expect(mobileNavButton).toBeInTheDocument();
+
+    // click the mobile nav button
+    await act(async () => {
+      await user.click(mobileNavButton);
+    });
+
+    // expect mobile nav to be visible
```
laurenrevere: Should the mobile nav not be visible before the button is clicked?

If so, it would be helpful to check that.

laurenrevere (approved):

laurenrevere:
Responding to your note about the download button, that seems like a good place to be for now! We might get more requirements around that during the sprint where we handle downloads.

packages/ui/src/header/Header.test.tsx
nicholaslyon-dt: Good call. Updated!

laurenrevere:
*Build*: dd9b06d1e1ef5131fbc78863973dc384eee500be
*Device Info:* Safari and Chrome on Sanoma 14.4.1, Edge and Chrome on Windows 11

Header at Web Size:
- Matches designs ⚠️
- Voiceover ✅ 
- Color contrast ✅ 
- Keyboard navigation is logical ⚠️

Header at Mobile Size:
- Matches designs
- Voiceover ✅ 
- Color contrast ✅ 
- Keyboard navigation is logical ⚠️ 



*Observations/Notes:*
1. We may want to be clearer about the fact that the download link will open in a new window. [Or don’t open a new window?](https://www.a11yproject.com/checklist/#use-list-elements-ol-ul-and-dl-elements-for-list-content)
2. Mobile Size: Keyboard navigation on a computer gets focused/stuck on the dialog when its expanded. Is a dialog the ideal solution for this? Voiceover also announces it as a dialog which may be confusing.
3. Should the BC logo be focusable? It currently doesn't link anywhere.
4. Header focus rings look different than the focus states of the rest of the app.
5. Designs show a "hover" state in the header. Do we want to replace that with the link hover states?


TODO: Review these screens on actual mobile.

PR #12, opened 2024-05-23
DT-1: Data transformation for 9.9.9.1

nicholaslyon-dt:
**Purpose**
Transform the Test Case SME document into JSON for use with the Walkthough

**Testing**
Review the Test Case document against the data in JSON to make sure content and logic is correct

**Note**
This is for all of 9.9.9.1.

### DIFF
--- a/packages/data/walkthroughs/9.9.9.json
+++ b/packages/data/walkthroughs/9.9.9.json
@@ -418,7 +418,7 @@
         },
         {
           "nextLogicType": "fallback",
-          "nextNavigateTo": "P05"
+          "nextNavigateTo": "R05"
         }
       ]
     },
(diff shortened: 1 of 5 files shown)


packages/data/walkthroughs/9.9.9.json
```diff
@@ -406,15 +412,1149 @@
...
+            },
+            {
+              "showAnswerLogicType": "equals",
+              "answerToCheck": "P2",
+              "answerValue": "2B"
+            },
+            {
+              "showAnswerLogicType": "equals",
+              "answerToCheck": "P2",
+              "answerValue": "3B"
```
laurenrevere, line 525: Can a building with more than 3 stories have a basement? Is that something we should be concerned about?

jbook24 (approved):

PR #13, opened 2024-05-24
FE-6: Initial CheckboxGroup work

nicholaslyon-dt:
**Purpose**
Create a reusable Checkbox component

**Testing**
Build the application and head to the homepage
See the example checkbox usage

### DIFF
--- a/packages/ui/src/checkbox-group/CheckboxGroup.test.tsx
+++ b/packages/ui/src/checkbox-group/CheckboxGroup.test.tsx
@@ -1,0 +1,318 @@
+// 3rd party
+import { describe, expect, it } from "vitest";
+import { act, render, waitFor } from "@testing-library/react";
+import { Form } from "react-aria-components";
+// repo
+import {
+  GET_TESTID_BUTTON,
+  GET_TESTID_CHECKBOX,
+  GET_TESTID_CHECKBOX_GROUP,
+  TESTID_BUTTON_SUBMIT,
+  TESTID_CHECKBOX_GROUP_ERROR,
+  TESTID_CHECKBOX_GROUP_LABEL,
+} from "@repo/constants/src/testids";
+// workspace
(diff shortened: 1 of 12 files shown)


laurenrevere:
*Build*: e041291c219eae994be3bca040c694ccde070971
*Device Info:* Safari and Chrome on Sanoma 14.4.1, Edge and Chrome on Windows 11

- Matches designs ⚠️
- Error state appearance ⚠️
- Voiceover ✅
- Error messages are announced when they appear ✅
- Color contrast ✅
- Keyboard navigation is logical ✅



*Observations/Notes:*
1. Only one checkbox should be focused at a time. Selected and focused should not be the same state. This also causes problems for being able to tell what it selected when navigating with screenreader or keyboard.
2. Error state should not have red text. Even in the error state the main text should remain black.


Screenshot 2024-05-28 at 2 16 31 PM

packages/ui/src/checkbox-group/CheckboxGroup.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    const input1 = option1.querySelector("input[type='checkbox']");
+    const input2 = option2.querySelector("input[type='checkbox']");
+    const input3 = option3.querySelector("input[type='checkbox']");
+    expect(input1).not.toBeChecked();
+    expect(input2).toBeChecked();
+    expect(input3).not.toBeChecked();
+  });
+  // selected value overrides default value
```
laurenrevere, line 143: Can more than one value be selected in a checkbox group by default? I would expect that they could since checkbox groups are multi select.

nicholaslyon-dt: yes they can! the value/defaultValue are in string array forms. ie ["1"] or ["1","2"]. That comes out of the box with the react-aria library.

PR #15, opened 2024-05-29
DT-2: Data transformation for 9.9.9.2

laurenrevere:
**Purpose**
Transform the Test Case SME document into JSON for use with the Walkthough

**Testing**
Review the Test Case document against the data in JSON to make sure content and logic is correct

**Note**
- It may make sense to use different stored values for the P11 answers. I use these values in P14. There I look for a value greater than 2 instead of just looking for the value 3 because it matches the solutions document better. Happy for other solutions though.
- I added all of the L3 and L4 logic to P17. I have a simplified evaluation that I'd be happy to share, but may want to share it after someone else has taken a look first.

This can be based off of `mian` once DT-1 is merged.

### DIFF
--- a/packages/data/walkthroughs/9.9.9.json
+++ b/packages/data/walkthroughs/9.9.9.json
@@ -1711,6 +1711,170 @@
           "nextNavigateTo": "P17"
         }
       ]
+    },
+    "P17": {
+      "walkthroughItemType": "multiChoice",
+      "questionText": "Does this <defined-term>building</defined-term> meet the definition of a <defined-term>secondary suite</defined-term>?",
+      "possibleAnswers": [
+        {
+          "answerDisplayText": "Yes",
+          "answerValue": "true"
+        },
+        {
+          "answerDisplayText": "No",
+          "answerValue": "false"
+        }
+      ],
+      "nextNavigationLogic": [
+        {
+          "nextLogicType": "and",
+          "valuesToCheck": [
+            {
+              "nextLogicType": "equals",
+              "answerToCheck": "P15",
+              "answerValue": "true"
+            },
+            {
+              "nextLogicType": "equals",
+              "answerToCheck": "P17",
+              "answerValue": "true"
+            }
+          ],
+          "nextNavigateTo": "P18"
+        },
+        {
+          "nextLogicType": "and",
+          "valuesToCheck": [
+            {
+              "nextLogicType": "equals",
+              "answerToCheck": "P15",
+              "answerValue": "true"
+            },
+            {
+              "nextLogicType": "equals",
+              "answerToCheck": "P17",
+              "answerValue": "false"
+            }
+          ],
+          "nextNavigateTo": "P20"
+        },
+        {
+          "nextLogicType": "and",
+          "valuesToCheck": [
+            {
+              "nextLogicType": "equal",
+              "answerToCheck": "P17",
+              "answerValue": "true"
+            },
+            {
+              "nextLogicType": "or",
+              "valuesToCheck": [
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.B",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.1",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.2",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.3",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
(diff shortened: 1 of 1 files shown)


packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,6 +1555,326 @@
...
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P8.3",
+                  "answerValue": "true"
+                }
+              ]
+            }
+          ],
```
laurenrevere, line 1814: Might make sense to change to just use the logic type "contains".

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,6 +1555,326 @@
...
+          "answerDisplayText": "1-2",
+          "answerValue": "2"
+        },
+        {
+          "answerDisplayText": "Greater than 2",
+          "answerValue": "3"
+        }
+      ],
+      "nextNavigationLogic": [
+        {
```
jbook24, line 1573: Should nextNavigationLogic include
   "answerToCheck": "P11" ?

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,6 +1555,326 @@
...
+          "answerValue": "false"
+        }
+      ],
+      "nextNavigationLogic": [
+        {
+          "nextLogicType": "and",
+          "valuesToCheck": [
+            {
+              "nextLogicType": "equals",
+              "answerToCheck": "P15",
```
jbook24, line 1734: Should this  "answerToCheck": "P15", be P17?

jbook24 (approved):

packages/data/walkthroughs/9.9.9.json
laurenrevere: I don't think so. This one only has fallback logic. It goes straight to P12. 

When Nick implements, the data will be stored so if p11 needs to be used it can be.

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,6 +1555,326 @@
...
+      ],
+      "nextNavigationLogic": [
+        {
+          "nextLogicType": "equals",
+          "answerToCheck": "P14",
+          "answerValue": "true",
+          "nextNavigateTo": "P17"
+        },
+        {
+          "nextLogicType": "greaterThan",
```
nicholaslyon-dt, line 1652: @laurenrevere Yeah to me this could go any which way but I like this solution since, as you said, it matches the SME document better. We really just need a value to key off of in a specific way.

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,6 +1555,326 @@
...
+        },
+        {
+          "nextLogicType": "fallback",
+          "nextNavigateTo": "P15"
+        }
+      ]
+    },
+    "P15": {
+      "walkthroughItemType": "multiChoice",
+      "questionText": "Does the egress door of this <defined-term>dwelling unit</defined-term> open onto a corridor that meets the requirements of Sentence 9.9.7.3.(1)?",
```
nicholaslyon-dt, line 1665: @laurenrevere We still don't have an exact solution for this case (link to out of scope Building Code area). My thoughts are:

Easy: Is a link that downloads the PDF.
Nice to have: Link contains a "named destination" like this link:
https://www2.gov.bc.ca/assets/gov/farming-natural-resources-and-industry/construction-industry/building-codes-and-standards/revisions-and-mo/bcbc_2024.pdf#nameddest=9.9.9.%20Egress%20from%20Dwelling%20Units

This would jump the user to that section of the document. But only in certain browsers and PDF readers that support this functionality.

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,6 +1555,326 @@
...
+        },
+        {
+          "nextLogicType": "fallback",
+          "nextNavigateTo": "P16"
+        }
+      ]
+    },
+    "P16": {
+      "walkthroughItemType": "multiChoice",
+      "questionText": "Is the <defined-term>building sprinklered</defined-term>?",
```
nicholaslyon-dt, line 1691: @laurenrevere This is technically 2 separate defined terms.

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,6 +1555,326 @@
...
+        {
+          "answerDisplayText": "Yes",
+          "answerValue": "true"
+        },
+        {
+          "answerDisplayText": "No",
+          "answerValue": "false"
+        }
+      ],
+      "nextNavigationLogic": [
```
nicholaslyon-dt, line 1728: @laurenrevere I'm not 100% following all the extra checks here from the previous questions that get pushed here. I may need a walk through for what you're thinking.

PR #16, opened 2024-05-30
DT-3: Data transformation for 9.9.9.3

jbook24:
View commit: **Initial transformation of 9.9.9.3** to view only changes related only to 9.9.9.3 

**Context**
- references DT-1: Data transformation for 9.9.9.1 
- Initial draft transformation of walkthrough 9.9.9.3 
- remove reference label "-remove-after-review-9.9.9.3" upon final review 
- **Do not merge** until Data transformation for 9.9.9.1 & DT-2: Data transformation for 9.9.9.2 are merged. 

**Tracking**
Commit: updates to p18-p20 (bug fixes)

### DIFF
--- a/packages/data/walkthroughs/9.9.9.json
+++ b/packages/data/walkthroughs/9.9.9.json
@@ -1688,7 +1688,7 @@
     },
     "P16": {
       "walkthroughItemType": "multiChoice",
-      "questionText": "Is the <defined-term>building sprinklered</defined-term>?",
+      "questionText": "Is the <defined-term>building</defined-term> <defined-term>sprinklered</defined-term>?",
       "possibleAnswers": [
         {
           "answerDisplayText": "Yes",
@@ -1726,6 +1726,12 @@
         }
       ],
       "nextNavigationLogic": [
+        {
+          "nextLogicType": "equals",
+          "answerToCheck": "P17",
+          "answerValue": "false",
+          "nextNavigateTo": "R20"
+        },
         {
           "nextLogicType": "and",
           "valuesToCheck": [
@@ -1904,7 +1910,7 @@
     },
     "P19": {
       "walkthroughItemType": "multiChoice",
-      "questionText": "Does the egress door of this <defined-term>dwelling unit</defined-term> open onto a shared egress facility (corridor, passageway, stair, ramp, etc.) that is served by a single stair or ramp?",
+      "questionText": "Does the egress door of this <defined-term>dwelling unit</defined-term> open onto a shared egress facility (corridor, passageway, stair, <defined-term>ramp</defined-term>, etc.) that is served by a single stair or <defined-term>ramp</defined-term>?",
       "possibleAnswers": [
         {
           "answerDisplayText": "Yes",
@@ -1925,25 +1931,49 @@
               "answerValue": "true"
             },
             {
-              "nextLogicType": "contains",
-              "answerToCheck": "P5",
-              "answerValue": "true"
-            }
-          ],
-          "nextNavigateTo": "R10"
-        },
-        {
-          "nextLogicType": "and",
-          "valuesToCheck": [
-            {
-              "nextLogicType": "equals",
-              "answerToCheck": "P19",
-              "answerValue": "true"
-            },
-            {
-              "nextLogicType": "contains",
-              "answerToCheck": "P8",
-              "answerValue": "true"
+              "nextLogicType": "or",
+              "valuesToCheck": [
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.B",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.1",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.2",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P5.3",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P8.B",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P8.1",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P8.2",
+                  "answerValue": "true"
+                },
+                {
+                  "nextLogicType": "equal",
+                  "answerToCheck": "P8.3",
+                  "answerValue": "true"
+                }
+              ]
             }
           ],
           "nextNavigateTo": "R10"
@@ -1956,7 +1986,7 @@
     },
     "P20": {
       "walkthroughItemType": "multiChoice",
-      "questionText": "Does the egress door of this <defined-term>dwelling unit</defined-term> open onto an exit stair, public corridor, exterior passageway, or balcony, that serves more than one suite?",
+      "questionText": "Does the egress door of this <defined-term>dwelling unit</defined-term> open onto an <defined-term>exit</defined-term> stair, <defined-term>public corridor</defined-term>, exterior passageway, or balcony, that serves more than one <defined-term>suite</defined-term>?",
       "possibleAnswers": [
         {
           "answerDisplayText": "Yes",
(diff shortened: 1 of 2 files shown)


packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,8 +1555,230 @@
...
+      "nextNavigationLogic": [
+        {
+          "nextLogicType": "equals",
+          "answerToCheck": "P23",
+          "answerValue": "true",
+          "nextNavigateTo": "P24"
+        },
+        {
+          "nextLogicType": "fallback",
+          "nextNavigateTo": "R1.0"
```
laurenrevere: This would need to be R10 as @nicholaslyon-dt doesn't include the "." in the results indices. The same would be true for other "R" values.

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1583,6 +1805,15 @@
...
     },
+    "R1.0-remove-after-review-9.9.9.3": {
+      "resultDisplayMessage": "The dwelling unit meets the requirement of sufficient means of egress as per BC Building Code 2024."
+    },
+    "R1.1-remove-after-review-9.9.9.3": {
+      "resultDisplayMessage": "The dwelling unit is located above another dwelling unit or common space. Specific requirements apply."
+    },
+    "R2.0-remove-after-review-9.9.9.3": {
+      "resultDisplayMessage": "The dwelling unit does not meet the requirement of sufficient means of egress as per BC Building Code 2024."
+    },
```
laurenrevere: I believe these can be removed. I think that they are redundant of R10, R11, and R20.

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,8 +1555,230 @@
           "nextNavigateTo": "P11"
         }
       ]
-    }
-  },
+    }, 
+    "P18-9.9.9.3-remove-after-review": {
+      "walkthroughItemType": "multiChoice",
+      "questionText": "Is this dwelling unit located above another dwelling unit or common space?",
```
laurenrevere: We need to add `<defined-term>` around italicized terms like "dwelling unit".

`<defined-term>dwelling unit</defined-term>`

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,8 +1555,230 @@
...
+        {
+          "nextLogicType": "and",
+          "valuesToCheck": [
+            {
+              "nextLogicType": "equals",
+              "answerToCheck": "P19",
+              "answerValue": "true"
+            },
+            {
+              "nextLogicType": "contains",
```
laurenrevere: Oooh, I don't see this exact logic style used anywhere yet in @nicholaslyon-dt's current data transformation, but would probably be reasonable to be part of his implementation.

In my implementation [I literally checked every value](https://github.com/bcgov/House-Innovations-HOUSNAV/pull/15/files#r1621209253), but this is easier to manage if it works. We can probably merge both, but just add it to the list of things that will need to be checked by Nick!

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1555,8 +1555,230 @@
...
+        },
+        {
+          "nextLogicType": "fallback",
+          "nextNavigateTo": "R2.0"
+        }
+      ]
+    },
+    "P23": {
+      "walkthroughItemType": "multiChoice",
+      "questionText": "Is the {answer from P21} served by a single exit stairway or ramp?",
```
laurenrevere: Note that @nicholaslyon-dt will have to decide a way to handle the variable in this question text.

laurenrevere:
I did end up validating against the doc. Looks good!

The results values will need to be changed and ideally the `<defined-term>` tags would be added in before merge.

Note: The variables in P23 and P24 will need to be sorted out before this full logic tree is usable by @nicholaslyon-dt.

packages/data/walkthroughs/9.9.9.json
nicholaslyon-dt: I would vote to re-write the data to conform to the "or" logic type with a bunch of "equal" checks for each value. It keeps the javascript more simple (in that we're not adding an extra type to manage/test) but I'm open to adding this if we feel like we definitely need it this way.

packages/data/walkthroughs/9.9.9.json
nicholaslyon-dt: Yeah I'd probably swap this with a custom tag with a data prop, something like `<answer-value answer="P21" />` and then add an extra property to the P21 answers like "answerValueDisplay": "exterior passageway".

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1726,6 +1726,12 @@
         }
       ],
       "nextNavigationLogic": [
+        {
+          "nextLogicType": "equals",
+          "answerToCheck": "P17",
+          "answerValue": "false",
+          "nextNavigateTo": "R20"
```
laurenrevere: I believe that this has to go after 

          "nextLogicType": "and",
          "valuesToCheck": [
            {
              "nextLogicType": "equals",
              "answerToCheck": "P15",
              "answerValue": "true"
            },
            {
              "nextLogicType": "equals",
              "answerToCheck": "P17",
              "answerValue": "false"
            }
          ],
          "nextNavigateTo": "P20"
        },

Otherwise, it won't handle this case correctly.

We need to ask the P15 questions first to account for the fact that P15(L11) goes straight to L14 if P15 is true.

laurenrevere: I am also realizing that I incorrectly added this case `If P5 or P8 contains true and p17 == false go to p20` and it can be removed.

packages/data/walkthroughs/9.9.9.json
```diff
@@ -1982,22 +2012,24 @@
...
           "answerValue": "A"
         },
         {
-          "answerDisplayText": "A public corridor",
+          "answerDisplayText": "A <defined-term>public corridor</defined-term>",
           "answerValue": "B"
         },
         {
           "answerDisplayText": "An exterior passageway",
+          "answerValueDisplay": "exterior passageway",
```
laurenrevere, line 1959: For my understanding, what's an `answerValueDisplay`?

packages/data/walkthroughs/9.9.9.json
laurenrevere: **Here's my new understanding of the logic:**

The only flow that goes to P14 directly is when P15=true. Because this needs to skip the L13 logic, we need to ask those first. Here's our  P15=true logic:

- If p15 = true and  p17 == true go to P18
- If p15 = true and  p17 == false go to P20


Then we can take L13 at face value.

> If P(17) == No/False go to R2.0
> Else if P(17) == True and P(5).contains(True) go to L14 
> Else if P(17) == True and P(8).contains(True) go to L14 
> Else go to R2.0


So we have: 

- If p15 = true and  p17 == true go to p18
- If p15 = true and  p17 == false go to p20
- If P(17) == No/False go to R2.0
- if P(17) == True and P(5).contains(True) go to L14 
- if P(17) == True and P(8).contains(True) go to L14 
- Else go to R2.0

Simplified to:
- If p15 = true and  p17 == true go to p18
- If p15 = true and  p17 == false go to p20
- If P(17) == No/False go to R2.0
- **If P(17) == True and (P5 or P8 contains(True)) go to L14** 
- Else go to R2.0

From here, we only navigate to the L14 logic if P17=True. So we can simplify again ignoring the "else" of L14:
- If p15 = true and  p17 == true go to p18
- If p15 = true and  p17 == false go to p20
- If P(17) == No/False go to R2.0
- **If P(17) == True and (P5 or P8 contains(True)) go to P18** 
- Else go to R2.0


Above could be your final answer, but you can simplify even further. You don't need to check if P17 is false, because the only other relevant option only checks if it's true.
- If p15 = true and  p17 == true go to p18
- Else If p15 = true and  p17 == false go to p20
- Else if P(17) == True and (P5 or P8 contains(True)) go to P18
- Else go to R2.0

nicholaslyon-dt: I walked through all the scenarios and got something slightly different I think? But it might be able to be simplified down to what you got. LMK what you think.

1. P17 === false && (P15 === undefined || P15 === false) ---> R20
2. P17 === false && P15 === true ---> P20
3. P17 === true && (P15 === true || P5/P8 contain true) ---> P18
4. else R20

After doing all this I'm trying to determine if we just have multiple version of P17 depending on how you get there to simplify this and keep it closer to the document. Thoughts?

packages/data/walkthroughs/9.9.9.json
nicholaslyon-dt: That is the display value for when that answer is used later in a question (for the question text fill in for P23 and P24). I thought about just making the `answerValue` also a display appropriate value but wasn't sure about a case where we'd want them different so opted for this. Thoughts?

packages/data/walkthroughs/9.9.9.json
laurenrevere: After discussing we realized the logic is the same and that it can be further simplified. Here's the final:

1. P17 === false && P15 === true ---> P20
2. P17 === true && (P15 === true || P5/P8 contain true) ---> P18
3. else R20

laurenrevere: Also I agree with your assessment that closer to the document is probably better in most cases so other people can understand our logic.

packages/data/walkthroughs/9.9.9.json
laurenrevere: Makes sense to me.

packages/data/walkthroughs/9.9.9.json
nicholaslyon-dt: Updated!

laurenrevere (approved):
Looks solid!

PR #17, opened 2024-05-30
HOUSNAV-11: Add unit tests to pipeline

laurenrevere:
Adds a unit test run to pipeline before building in docker.

Builds on all PRs as well as on main.


clay-comer-wt (approved):
Nice! LGTM

PR #19, opened 2024-05-30
HOUSNAV-11: Block docker build on unit test failure

laurenrevere:


laurenrevere:
Opened to try to revert merge instead of rebase on main. Handled separately.

PR #21, opened 2024-06-04
HOUSNAV-45: adding code references to walkthrough 9.9.9

nicholaslyon-dt:
**Purpose**
Adding JSON data for each question for the number code of the building code that is references. This is the data needed to accomplish this part of the design for each question:
Screenshot 2024-06-04 at 2 28 35 PM

**Notes**
1. Not all questions will have this, since the intro set of question (P01-P08) don't fully encompass a numbered part of the building code and the SME didn't feel comfortable pointing to one.
2. Actual functionality will come later

### DIFF
--- a/apps/web/components/question/Question.test.tsx
+++ b/apps/web/components/question/Question.test.tsx
@@ -117,7 +117,10 @@ describe("Question", () => {
 
     // add questionCodeReference if it does not exist in questionData
     if (!questionInfo.questionData.questionCodeReference) {
-      questionInfo.questionData.questionCodeReference = "code reference";
+      questionInfo.questionData.questionCodeReference = {
+        displayString: "code reference",
+        codeNumber: "1",
+      };
     }
 
     // mock setQuestion function and render component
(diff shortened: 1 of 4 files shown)


laurenrevere (approved):

apps/web/components/question/Question.test.tsx
```diff
@@ -117,7 +117,10 @@ describe("Question", () => {
 
     // add questionCodeReference if it does not exist in questionData
     if (!questionInfo.questionData.questionCodeReference) {
-      questionInfo.questionData.questionCodeReference = "code reference";
+      questionInfo.questionData.questionCodeReference = {
+        displayString: "code reference",
+        codeNumber: "1",
```
laurenrevere, line 122: Should we check the value of the code reference in the test instead of just that it's there?

nicholaslyon-dt: Good call. Added that expect!

jbook24 (approved):

PR #22, opened 2024-06-04
HOUSNAV-44: adding sections object to walkthrough 9.9.9

nicholaslyon-dt:
**Purpose**
Adding a top level object that denotes the sections of the walkthrough, their titles, and which questions are in each section. This will help construct the side navigation in the future.
I've also added tests to verify 2 things:
Each question that appears in a section actually exists
Each question either appears in a section or is of type "variable"


apps/web/components/walkthrough/Walkthrough.tsx
```diff
@@ -37,12 +39,25 @@ export default function Walkthrough({
     // setVariable(currentQuestion.variableToSet);
```
jbook24: Do we want to remove this commented code?

jbook24 (approved):

apps/web/components/walkthrough/Walkthrough.tsx
nicholaslyon-dt: Good call. Removed!

packages/data/walkthroughs/9.9.9.test.ts
```diff
@@ -0,0 +1,46 @@
...
+          return false;
+        }
+
+        return !Object.values(data.sections).some((section) =>
+          section.sectionQuestions.includes(questionId),
+        );
+      },
+    );
+    if (possibleUnusedQuestion)
+      console.log("Possible unused question ID:", possibleUnusedQuestion[0]);
```
laurenrevere, line 22: Do we need this console log in the test? Won't the expect tell us the value of `possibleUnusedQuestion` if the test fails?

wardm89 (approved):
LGTM; did some some manual tests to verify it updates correctly to the next question.

laurenrevere (approved):

packages/data/walkthroughs/9.9.9.test.ts
nicholaslyon-dt: It does, but it's not as easy to read/readily apparent IMO. This was just for readability.

PR #24, opened 2024-06-05
HOUSNAV-53: Add PR template

wardm89:
[HOUSNAV-53](https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3?selectedIssue=HOUSNAV-53)

## Overview & Purpose

Add PR template to reduce time writing PR descriptions and standardizing PRs enabling faster testing times and "future selves" looking at old PRs to see what changed.

## Summary of Changes

- Adding template to the `.github` directory

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,18 @@
+<!-- You may remove any sections that are not applicable -->
+
+[HOUSNAV-#](https://hous-bssb.atlassian.net/browse/HOUSNAV-#)
+
+## Overview & Purpose
+<!-- Please describe the purpose of your changes. Please also include relevant motivation and context. -->
+
+## Summary of Changes
+<!-- Please include a HIGH LEVEL overview of the code changes. (Added xyz fields to table, modified function to handle x case instead of y case, etc. )  -->
+
+## Side Effects
+<!-- Any possible side effects? Does this change affect features that are not linked to the direct purpose? -->
+
+## Testing
+<!-- Please include steps to test out the changes/fixes if applicable OR context notes to help get env in a state to test-->
+
+## Notes & Resources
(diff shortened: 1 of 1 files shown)


nicholaslyon-dt (approved):

nicholaslyon-dt (changes requested):
Actually, can we get the ticket helper to have a default that matches our JIRA? Make it quicker to fill in?

`[HOUSNAV-10](https://hous-bssb.atlassian.net/browse/HOUSNAV-10)`

nicholaslyon-dt:
Ooo we should also add a question for Testing Notes.

wardm89:
> Ooo we should also add a question for Testing Notes.

The `Notes and Resources` is there any testing notes can go. However I'm fine with breaking that out into a separate section if we really want that additional distinction.

laurenrevere (approved):
Looks good to me! I do think we could make this even shorter if we wanted. 

I don't think we need a specific testing section, but if we did add one I'd combine side effects with notes and add testing at the end.


Update: but @nicholaslyon-dt's format shown here probably will work well as well. https://github.com/bcgov/House-Innovations-HOUSNAV/pull/25

nicholaslyon-dt:
> > Ooo we should also add a question for Testing Notes.
> 
> The `Notes and Resources` is there any testing notes can go. However I'm fine with breaking that out into a separate section if we really want that additional distinction.

I would appreciate breaking it out as I tend to always need to list testing steps and those would just overtake any notes I add. I would think separate is easier to read and write:

Screenshot 2024-06-05 at 1 11 31 PM

Than together:

Screenshot 2024-06-05 at 1 12 19 PM

Thoughts y'all?

wardm89:
I'm good with this as it stands right now. If we come up with any other changes down the line we can update it then.

One thing to note is all the sections aren't necessary so if one doesn't apply just delete it.

nicholaslyon-dt (approved):

PR #25, opened 2024-06-05
HOUSNAV-25: Multi-choice multiple question module

nicholaslyon-dt:
[HOUSNAV-25](https://hous-bssb.atlassian.net/browse/HOUSNAV-25)

## Overview & Purpose
To build the component for the multi choice multiple selection question module

## Summary of Changes
Update the `QuestionMultiChoiceMultiple` to use the new `CheckboxGroup` component
Updated temp state functions to be able to navigate between the first 3 questions to show them

## Side Effects
I fixed 2 bugs in the CheckboxGroup and RadioGroup components that I found
1. Parsed components in text strings weren't set up appropriately
2. Icons were shrinking if the text was wrapping

## Testing
1. Navigate to `Walkthrough 9.9.9`
2. Answer question 1 (any answer will do)
3. Verify question 2 displays with new multichoice multiple selection question
4. Answer question 2 (any answer will do)
5. Verify question 3 displays with new multichoice multiple selection question

## Notes & Resources
I have plans to redo state with the [Mobx React Lite implementation](https://mobx.js.org/react-integration.html) so don't spend too much time reviewing state updates.

### DIFF
--- a/apps/web/components/question/Question.tsx
+++ b/apps/web/components/question/Question.tsx
@@ -1,5 +1,5 @@
 // 3rd party
-import { FormEvent, useCallback, useState } from "react";
+import { FormEvent, useCallback } from "react";
 import { Form } from "react-aria-components";
 // repo
 import {
@@ -21,17 +21,25 @@ import Button from "@repo/ui/button";
 import QuestionMultiChoice from "./question-types/QuestionMultiChoice";
 import QuestionMultiChoiceMultiple from "./question-types/QuestionMultiChoiceMultiple";
 import { parseStringToComponents } from "../../utils/string";
+import {
+  NavigateToNextQuestionFunction,
+  SetAnswerValueFunction,
+  WalkthroughAnswersStateType,
+  WalkthroughAnswerType,
+} from "../walkthrough/Walkthrough";
 import "./Question.css";
 
 interface QuestionProps {
+  walkthroughAnswersState: WalkthroughAnswersStateType;
   questionData: QuestionDisplayData;
   questionId: string;
-  setNextQuestion: (question: string) => void;
+  navigateToNextQuestion: NavigateToNextQuestionFunction;
+  setAnswerValue: SetAnswerValueFunction;
 }
 
 export interface SharedQuestionProps {
-  value?: string;
-  setValue: (value: string) => void;
+  value?: WalkthroughAnswerType;
+  setValue: SetAnswerValueFunction;
   questionId: string;
 }
 
@@ -49,7 +57,7 @@ const getQuestionComponent = ({
   if (isWalkthroughItemTypeMultiChoice(questionData.walkthroughItemType)) {
     return (
       <QuestionMultiChoice
-        value={value}
+        value={value as string}
         setValue={setValue}
         questionId={questionId}
         {...(questionData as QuestionMultipleChoiceData)}
@@ -60,7 +68,7 @@ const getQuestionComponent = ({
   ) {
     return (
       <QuestionMultiChoiceMultiple
-        value={value}
+        value={value as string[]}
         setValue={setValue}
         questionId={questionId}
         {...(questionData as QuestionMultipleChoiceSelectMultipleData)}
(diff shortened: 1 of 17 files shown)


apps/web/components/walkthrough/Walkthrough.tsx
```diff
@@ -17,11 +19,32 @@ interface WalkthroughProps {
   startingSectionId: StartingSectionIdType;
 }
 
+export type WalkthroughAnswerType =
+  | string
+  | string[]
+  | Record<string, string>
+  | undefined;
```
wardm89, line 26: Do we want to actually handle undefined here with the answer type since an undefined answer doesn't make sense we could handle it at a higher level? 

We may have to do this (handle undefined in this type); not sure right now with my understanding of the code.

packages/ui/src/checkbox-group/CheckboxGroup.test.tsx
```diff
@@ -30,12 +30,7 @@ const CHECKBOX_GROUP_OPTIONS = [
   { label: "1 storey", value: OPTION_1_VALUE },
   { label: "2 storeys", value: OPTION_2_VALUE },
   {
-    label: (
-      <>
-        3&nbsp;
-        <Button variant={"glossary"}>storeys</Button>
-      </>
-    ),
+    label: '3 <Button variant="glossary">storeys</Button>',
```
wardm89, line 33:
```suggestion
    label: '3 <Button variant="glossary">storeys</Button>',
```
```suggestion
    label: '3 <Button variant="glossary">stories</Button>',
```
Typo?

packages/ui/src/radio-group/RadioGroup.test.tsx
```diff
@@ -30,12 +30,7 @@ const RADIO_GROUP_OPTIONS = [
   { label: "1 storey", value: OPTION_1_VALUE },
   { label: "2 storeys", value: OPTION_2_VALUE },
   {
-    label: (
-      <>
-        3&nbsp;
-        <Button variant={"glossary"}>storeys</Button>
-      </>
-    ),
+    label: '3 <Button variant="glossary">storeys</Button>',
```
wardm89, line 33: typo?

apps/web/components/walkthrough/Walkthrough.tsx
nicholaslyon-dt: Yeah we'll see. It's currently because of how I'm tracking state but since I'm going to redo that part at a higher level I was waiting to see what comes out in the wash.

packages/ui/src/radio-group/RadioGroup.test.tsx
nicholaslyon-dt: No sadly, that's the Canadian way they spell story in the Building Code so I decided to keep the example as such.

jbook24:
@nicholaslyon-dt when navigating from question to question - are you experiencing the scroll not resetting to the top of the screen?

apps/web/components/question/Question.tsx
```diff
@@ -49,7 +57,7 @@ const getQuestionComponent = ({
   if (isWalkthroughItemTypeMultiChoice(questionData.walkthroughItemType)) {
     return (
       <QuestionMultiChoice
-        value={value}
+        value={value as string}
```
jbook24, line 60: Do we anticipate this being a persistent pattern? Curious if there's a better way to handle this mutation?

packages/ui/src/radio-group/RadioGroup.css
jbook24, line 1: This code seems duplicative of the checkboxGroup.css 
Should we created a shared implementation since they appear to be the same? 
.ui-Checkbox--Icon {.. }

apps/web/components/question/Question.tsx
nicholaslyon-dt: I don't think. I'm hoping I don't have to do it once I redo state at a higher level so I have to pass less things around and everything can just subscribe to what it needs. I'm starting that migration with the footer ticket.

nicholaslyon-dt:
> @nicholaslyon-dt when navigating from question to question - are you experiencing the scroll not resetting to the top of the screen?

I am. I've added it as an AC on the footer ticket https://hous-bssb.atlassian.net/browse/HOUSNAV-22 as it relates more to that functionality of going between questions.

packages/ui/src/radio-group/RadioGroup.css
nicholaslyon-dt: I thought about it when I did the CheckboxGroup component, but I didn't want to optimize too soon. The code is easier to understand with them separate as well. I'm open to changing it thought. Should we create a backlog refinement ticket for this possible optimization?

PR #34, opened 2024-06-12
HOUSNAV-19: Side Modal Skeleton

wardm89:
[HOUSNAV-19](https://hous-bssb.atlassian.net/browse/HOUSNAV-19)

## Overview & Purpose

Add Modal popup on side of screen when user clicks a defined term. 

## Summary of Changes

- Adds `ModalSide` component
- Implements `ModalSide` component for all defined terms
- Adds `definedTermModal` case when parsing components to prevent infinite loop of opening modals and enable scrolling to sections within the modal when open
- Adds `--theme-black-opacity-40`
- Adds dummy data for how we can potentially implement this modal

## Testing

- Click any defined term. 
    - Note the sections that the terms navigate to are randomly generated right now 
  
Screenshot 2024-06-12 at 12 35 59 PM

https://github.com/bcgov/House-Innovations-HOUSNAV/assets/7059124/ef51b20f-5eb2-45a0-a97a-2b9922a8f072

## Notes & Resources

- This ticket didn't technically include css formatting for the data in the modal so it's currently a "rough draft" for the CSS.
    - It should be completed with this ticket: https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3/backlog?selectedIssue=HOUSNAV-71 
https://react-spectrum.adobe.com/react-aria/Modal.html#modal

### DIFF
--- a/apps/web/components/defined-term/DefinedTerm.tsx
+++ b/apps/web/components/defined-term/DefinedTerm.tsx
@@ -1,7 +1,9 @@
 // repo
 import Button, { ButtonProps } from "@repo/ui/button";
 import { TESTID_DEFINED_TERM } from "@repo/constants/src/testids";
+import { GLOSSARY_SECTIONS, GLOSSARY_TERMS } from "../../../web/tests/mockData";
 import Tooltip from "@repo/ui/tooltip";
+import ModalSide from "@repo/ui/modal-side";
 
 export interface DefinedTermProps extends Omit<ButtonProps, "variant"> {
   term: string;
@@ -14,23 +16,32 @@ export default function DefinedTerm({
   ...props
 }: DefinedTermProps) {
   // TODO: Add glossary term functionality
-  const mockGlossary = {
-    default: { tooltipContent: "Default Tooltip" },
-  };
 
-  return (
+  const button = (
+    <Button
+      variant="glossary"
+      data-term={term}
+      data-testid={`${testid}-${term}`}
+      {...props}
+    >
+      {children}
+    </Button>
+  );
+
+  const tooltipButton = (
     <Tooltip
-      tooltipContent={mockGlossary.default.tooltipContent}
-      triggerContent={
-        <Button
-          variant="glossary"
-          data-term={term}
-          data-testid={`${testid}-${term}`}
-          {...props}
-        >
-          {children}
-        </Button>
-      }
+      tooltipContent={GLOSSARY_TERMS.default.tooltipContent}
+      triggerContent={button}
     ></Tooltip>
+  );
+
+  // TODO: Matt or Nicholas, add correct section for customHandler
+  const randomSection = Math.floor(Math.random() * 10) + 1;
+  return (
+    <ModalSide
+      triggerContent={tooltipButton}
+      sections={GLOSSARY_SECTIONS}
+      scrollToSection={`9.9.9.${randomSection}`}
+    />
   );
 }
--- a/apps/web/utils/string.tsx
+++ b/apps/web/utils/string.tsx
@@ -10,19 +10,22 @@ import parse, {
 import DefinedTerm, {
   DefinedTermProps,
 } from "../components/defined-term/DefinedTerm";
+import Button from "@repo/ui/button";
 
 // Define custom components for html-react-parser
 const definedTermName = "defined-term";
-type CustomComponentTypes = typeof definedTermName;
+const definedTermModal = "defined-term-modal";
+type CustomComponentTypes = typeof definedTermName | typeof definedTermModal;
 type CustomComponentProps = DefinedTermProps;
 const customComponents: Record<
   CustomComponentTypes,
   FunctionComponent<CustomComponentProps>
 > = {
   [definedTermName]: DefinedTerm,
+  [definedTermModal]: Button,
 };
 
-export const parseStringToComponents = (html: string) => {
+export const parseStringToComponents = (html: string, customHandler?: any) => {
   const options = {
     replace: (domNode: DOMNode) => {
       if (domNode instanceof Element && domNode.attribs) {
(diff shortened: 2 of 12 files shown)


wardm89:
Note: This ticket didn't technically include css formatting (we discussed just having a white/empty popup) for the data in the modal so it's currently a "rough draft" for the CSS. I just wanted to get things in a good spot and wanted to test out the "scroll to section" functionality which required at least a little CSS to get it to make sense.
    - It should be completed with this ticket I believe: https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3/backlog?selectedIssue=HOUSNAV-71

jbook24:
Are we thinking the formatting cut off text at the bottom would be fixed in HOUSNAV-71 as well? 

image

apps/web/components/defined-term/DefinedTerm.tsx
```diff
@@ -14,23 +16,32 @@ export default function DefinedTerm({
...
-        >
-          {children}
-        </Button>
-      }
+      tooltipContent={GLOSSARY_TERMS.default.tooltipContent}
+      triggerContent={button}
     ></Tooltip>
   );
+
+  // TODO: Matt or Nicholas, add correct section for customHandler
```
jbook24, line 38: Should we link the ticket this should be addressed in?

apps/web/utils/string.tsx
```diff
@@ -42,6 +47,27 @@ export const parseStringToComponents = (html: string) => {
               );
             }
           }
+          case definedTermModal: {
+            // TODO: Matt or Nicholas, add correct section for customHandler
```
jbook24, line 53: Same here?

packages/ui/src/modal-side/ModalSide.css
```diff
@@ -0,0 +1,141 @@
...
+  display: flex;
+  flex-direction: column;
+  padding: var(--layout-padding-large);
+
+  &[data-entering] {
+    animation: modal-slide-in 300ms reverse ease-in;
+  }
+
+  &[data-exiting] {
+    animation: modal-slide-out 300ms reverse ease-in;
```
jbook24, line 38: Curious to understand how this animation performs on slower machines? Do we think a 

will-change: auto is needed?

packages/ui/src/modal-side/ModalSide.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <ModalOverlay className="ui-ModalSide--Overlay">
+        <Modal
+          className="ui-ModalSide--Modal"
+          ref={modalRef}
+          data-testid={testid}
+        >
+          <Dialog className="ui-ModalSide--AriaDialog">
+            {({ close }) => (
+              <>
+                <div className="ui-ModalSide--Header">
```
jbook24: Is there an opportunity to use semantic Html5 within the Modal? 
```

<header> 

<section> 

<div>

<aside>

```
etc..

apps/web/components/defined-term/DefinedTerm.tsx
wardm89: Idts, I'm not sure exactly which ticket it will be but it will definitely be covered when the terms are added and the data structure is implemented.

packages/ui/src/modal-side/ModalSide.css
wardm89: `Warning: will-change is intended to be used as a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems.`
https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

packages/ui/src/modal-side/ModalSide.tsx
wardm89: done

wardm89:
> Are we thinking the formatting cut off text at the bottom would be fixed in HOUSNAV-71 as well?
> 
> image

You should be able to scroll. (See test video) 
If scrolling doesn't appear then stop and restart the app. If it still doesn't appear what browser are you using and are any errors occuring?

jbook24 (approved):

packages/ui/src/modal-side/ModalSide.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    isIconButton
+                    variant="tertiary"
+                    className="ui-ModalSide--CloseButton"
+                    onPress={close}
+                    data-testid={`${testid}-close-button`}
+                  >
+                    <Icon type="close" />
+                  </Button>
+                </header>
+                <main className="ui-ModalSide--Content">
```
nicholaslyon-dt: I like the semantic updates for a lot of this, but there is already a main on the page and there can only be one main element on a page semantically unless the other one is hidden. Ours is still in the DOM below the modal.

packages/ui/src/variables.css
```diff
@@ -263,3 +263,8 @@
     --question-container-max-width: 53.5rem; /* 856px */
   }
 }
+
+/* CUSTOM */
+:root {
+  --theme-black-opacity-40: #00000040;
```
nicholaslyon-dt, line 269: I've just been putting custom ones at the bottom of the definitions above. I don't mind the breakout here if we want to refactor that to separate them. However, if they are specific to a component I've been trying to namespace them. I could see this being named `--modal-side-overlay-background-color`. Thoughts?

apps/web/utils/string.tsx
```diff
@@ -42,6 +47,27 @@ export const parseStringToComponents = (html: string) => {
               );
             }
```
nicholaslyon-dt, line 50: This is my error but since there was only 1 case it wasn't an issue. I'm getting a warning that we need some `break;`s at the end of our cases now.

nicholaslyon-dt (changes requested):
Left a couple thoughts and fixes. LMK what you think!

apps/web/utils/string.tsx
wardm89: Yeah I see, I missed that they didn't have breaks good catch. I think that's another product of my vscode env wanting to get all it's linting checks from what's in the vscode folder. In other TS repos I have it picks up on those things. 

I may make some more changes to the vscode settings.json to resolve that.

packages/ui/src/variables.css
wardm89: Right now it's technically specific to this component but down the line if we add more modals in we can just use. 

Also, I do like the idea of breaking it out into a "custom" section to separate them from the "defaults" that shouldn't be adjusted.

nicholaslyon-dt: Cool. Do you want to leave a todo note here for a possible code quality refactor later or capture that in a ticket? Not sure on the best way to handle refactoring what's there already.

nicholaslyon-dt (approved):

PR #35, opened 2024-06-13
Housnav-21: Walkthrough Card Component

wardm89:
[HOUSNAV-21](https://hous-bssb.atlassian.net/browse/HOUSNAV-21)

## Overview & Purpose

Add a reusable walkthrough card component to be used through the app. (Home page and end result)

## Summary of Changes

- Adds `walkthoughCard` component
- Updates the `9.9.9.` walkthrough to add titles and descriptions for the cards
    - Updates the `WalkthroughInfo` accordingly 

## Side Effects

- Adjusted `useWalkthroughData` error handling to make it more understandable

## Testing

- Go to home screen and new card should appear
- It should navigate to the 9.9.9 walkthrough when clicked

https://github.com/bcgov/House-Innovations-HOUSNAV/assets/7059124/52b6de35-7346-4a12-9226-49f6de218863

Screenshot 2024-06-13 at 9 59 12 AM

Screenshot 2024-06-13 at 9 59 30 AM

### DIFF
--- a/apps/web/app/page.tsx
+++ b/apps/web/app/page.tsx
@@ -10,21 +10,24 @@ import {
 
 export default function Page(): JSX.Element {
   return (
-    <div className="u-container ">
-      {Walkthroughs.map((walkthrough: string) => {
-        const data = WalkthroughJSONData[walkthrough]?.info;
-        const canDisplayCard =
-          data && data.title && data.subtitle && data.description;
-        return (
-          canDisplayCard && (
-            <WalkthroughCard
-              id={walkthrough}
-              data={data}
-              href={`${URL_WALKTHROUGH_HREF}/${walkthrough}/`}
-            />
-          )
-        );
-      })}
+    <div className="u-container">
+      <div>
+        {Object.entries(WalkthroughJSONData).map(([id, data]) => {
+          const info = data.info;
+          const canDisplayCard =
+            info && info.title && info.subtitle && info.description;
+          return (
+            canDisplayCard && (
+              <WalkthroughCard
+                key={id}
+                id={id}
+                data={info}
+                href={`${URL_WALKTHROUGH_HREF}/${id}/`}
+              />
+            )
+          );
+        })}
+      </div>
     </div>
   );
 }
--- a/apps/web/app/page.tsx
+++ b/apps/web/app/page.tsx
@@ -9,9 +9,7 @@ export default function Page(): JSX.Element {
     <div className="u-container">
       <div>
         {Object.entries(WalkthroughJSONData).map(([id, { info }]) => {
-          return (
-            <WalkthroughCard key={id} id={id} data={info} walkthroughId={id} />
-          );
+          return <WalkthroughCard key={id} data={info} walkthroughId={id} />;
         })}
       </div>
     </div>
--- a/apps/web/app/page.tsx
+++ b/apps/web/app/page.tsx
@@ -1,9 +1,8 @@
 // 3rd party
 import { JSX } from "react";
 // repo
-import { URL_WALKTHROUGH_HREF } from "@repo/constants/src/urls";
 import WalkthroughCard from "@repo/ui/walkthrough-card";
-import { WalkthroughJSONData } from "../../../packages/data/src/useWalkthroughData";
+import { WalkthroughJSONData } from "@repo/data/useWalkthroughData";
 
 export default function Page(): JSX.Element {
   return (
@@ -11,11 +10,7 @@ export default function Page(): JSX.Element {
       <div>
         {Object.entries(WalkthroughJSONData).map(([id, { info }]) => {
           return (
-            <WalkthroughCard
-              id={id}
-              data={info}
-              href={`${URL_WALKTHROUGH_HREF}/${id}/`}
-            />
+            <WalkthroughCard key={id} id={id} data={info} walkthroughId={id} />
           );
         })}
       </div>
--- a/apps/web/app/page.tsx
+++ b/apps/web/app/page.tsx
@@ -1,15 +1,30 @@
 // 3rd party
 import { JSX } from "react";
 // repo
-import Link from "@repo/ui/link";
 import { URL_WALKTHROUGH_HREF } from "@repo/constants/src/urls";
+import WalkthroughCard from "@repo/ui/walkthrough-card";
+import {
+  Walkthroughs,
+  WalkthroughJSONData,
+} from "../../../packages/data/src/useWalkthroughData";
 
 export default function Page(): JSX.Element {
   return (
-    <div className="u-container">
-      <p>
-        <Link href={`${URL_WALKTHROUGH_HREF}/9.9.9/`}>Walkthrough 9.9.9</Link>
-      </p>
+    <div className="u-container ">
+      {Walkthroughs.map((walkthrough: string) => {
+        const data = WalkthroughJSONData[walkthrough]?.info;
+        const canDisplayCard =
+          data && data.title && data.subtitle && data.description;
+        return (
+          canDisplayCard && (
+            <WalkthroughCard
+              id={walkthrough}
+              data={data}
+              href={`${URL_WALKTHROUGH_HREF}/${walkthrough}/`}
+            />
+          )
+        );
+      })}
     </div>
   );
 }
--- a/apps/web/app/page.tsx
+++ b/apps/web/app/page.tsx
@@ -3,28 +3,19 @@ import { JSX } from "react";
 // repo
 import { URL_WALKTHROUGH_HREF } from "@repo/constants/src/urls";
 import WalkthroughCard from "@repo/ui/walkthrough-card";
-import {
-  Walkthroughs,
-  WalkthroughJSONData,
-} from "../../../packages/data/src/useWalkthroughData";
+import { WalkthroughJSONData } from "../../../packages/data/src/useWalkthroughData";
 
 export default function Page(): JSX.Element {
   return (
     <div className="u-container">
       <div>
-        {Object.entries(WalkthroughJSONData).map(([id, data]) => {
-          const info = data.info;
-          const canDisplayCard =
-            info && info.title && info.subtitle && info.description;
+        {Object.entries(WalkthroughJSONData).map(([id, { info }]) => {
           return (
-            canDisplayCard && (
-              <WalkthroughCard
-                key={id}
-                id={id}
-                data={info}
-                href={`${URL_WALKTHROUGH_HREF}/${id}/`}
-              />
-            )
+            <WalkthroughCard
+              id={id}
+              data={info}
+              href={`${URL_WALKTHROUGH_HREF}/${id}/`}
+            />
           );
         })}
       </div>
--- a/packages/data/src/useWalkthroughData.ts
+++ b/packages/data/src/useWalkthroughData.ts
@@ -1,6 +1,6 @@
 import testCase999 from "../walkthroughs/9.9.9.json";
 
-export const Walkthroughs = ["9.9.9"];
+type Walkthroughs = "9.9.9";
 
 interface UseWalkthroughDataProps {
   /*
@@ -162,7 +162,7 @@ export interface WalkthroughJSONType {
   };
 }
 
-export const WalkthroughJSONData: Record<string, WalkthroughJSONType> = {
+export const WalkthroughJSONData: Record<Walkthroughs, WalkthroughJSONType> = {
   "9.9.9": testCase999,
 };
 
@@ -173,7 +173,7 @@ export default function useWalkthroughData({
     throw new Error("No id provided");
   }
 
-  const data = WalkthroughJSONData[id];
+  const data = WalkthroughJSONData[id as Walkthroughs];
   if (!data) {
     throw new Error(`No data found for walkthrough ${id}`);
   }
--- a/packages/data/src/useWalkthroughData.ts
+++ b/packages/data/src/useWalkthroughData.ts
@@ -108,7 +108,7 @@ export interface QuestionMultipleChoiceData extends QuestionBaseData {
 
 export const WalkthroughItemTypeMultiChoiceMultiple = "multiChoiceMultiple";
 export const isWalkthroughItemTypeMultiChoiceMultiple = (
-  walkthroughItemType: string
+  walkthroughItemType: string,
 ) => walkthroughItemType === WalkthroughItemTypeMultiChoiceMultiple;
 export interface QuestionMultipleChoiceSelectMultipleData
   extends QuestionBaseData {
@@ -169,10 +169,6 @@ export const WalkthroughJSONData: Record<Walkthroughs, WalkthroughJSONType> = {
 export default function useWalkthroughData({
   id,
 }: UseWalkthroughDataProps): WalkthroughJSONType {
-  if (!id) {
-    throw new Error("No id provided");
-  }
-
   const data = WalkthroughJSONData[id as Walkthroughs];
   if (!data) {
     throw new Error(`No data found for walkthrough ${id}`);
--- a/packages/data/src/useWalkthroughData.ts
+++ b/packages/data/src/useWalkthroughData.ts
@@ -1,6 +1,6 @@
 import testCase999 from "../walkthroughs/9.9.9.json";
 
-type Walkthroughs = "9.9.9";
+export const Walkthroughs = ["9.9.9"];
 
 interface UseWalkthroughDataProps {
   /*
@@ -108,7 +108,7 @@ export interface QuestionMultipleChoiceData extends QuestionBaseData {
 
 export const WalkthroughItemTypeMultiChoiceMultiple = "multiChoiceMultiple";
 export const isWalkthroughItemTypeMultiChoiceMultiple = (
-  walkthroughItemType: string,
+  walkthroughItemType: string
 ) => walkthroughItemType === WalkthroughItemTypeMultiChoiceMultiple;
 export interface QuestionMultipleChoiceSelectMultipleData
   extends QuestionBaseData {
@@ -138,15 +138,19 @@ interface SectionData {
   sectionQuestions: string[];
 }
 
+export interface WalkthroughInfo {
+  title: string;
+  subtitle: string;
+  description: string;
+  startingSectionId: StartingSectionIdType;
+}
+
 export type StartingSectionIdType = string;
 export type QuestionDisplayData =
   | QuestionMultipleChoiceData
   | QuestionMultipleChoiceSelectMultipleData;
 export interface WalkthroughJSONType {
-  info: {
-    title: string;
-    startingSectionId: StartingSectionIdType;
-  };
+  info: WalkthroughInfo;
   sections: {
     [key: string]: SectionData;
   };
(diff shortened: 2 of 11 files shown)


jbook24:
Could we apply an ease-in-out transition on hover? maybe 300ms?
https://stackoverflow.com/questions/16163922/css3-transition-ease-in-and-out-box-shadow

apps/web/app/page.tsx
```diff
@@ -1,15 +1,24 @@
...
 import { JSX } from "react";
 // repo
-import Link from "@repo/ui/link";
 import { URL_WALKTHROUGH_HREF } from "@repo/constants/src/urls";
+import WalkthroughCard from "@repo/ui/walkthrough-card";
+import { WalkthroughJSONData } from "../../../packages/data/src/useWalkthroughData";
 
 export default function Page(): JSX.Element {
   return (
     <div className="u-container">
```
jbook24, line 9: Should this div be a section? 
```
<section> 
   <div>
```

apps/web/app/page.tsx
```diff
@@ -1,15 +1,24 @@
...
-      <p>
-        <Link href={`${URL_WALKTHROUGH_HREF}/9.9.9/`}>Walkthrough 9.9.9</Link>
-      </p>
+      <div>
```
jbook24, line 10: Same here

apps/web/app/page.tsx
wardm89: The landing page html will be done with https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3/backlog?selectedIssue=HOUSNAV-30

wardm89:
> Could we apply an ease-in-out transition on hover? maybe 300ms? https://stackoverflow.com/questions/16163922/css3-transition-ease-in-and-out-box-shadow

Added a transition to the shadow and underline

jbook24 (approved):

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+    if (href) router.push(href);
+  };
+
+  const canDisplayCard =
+    data && data.title && data.subtitle && data.description;
+
+  if (!canDisplayCard) return null;
+
+  return (
+    <Button
```
nicholaslyon-dt: I think semantically this should be an `<a>` tag and not a button as it's sole purpose is to navigate via an href.

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+      className="ui-WalkthroughCard--CardContainer"
+      onPress={handleClick}
+      aria-label={`${id}-walkthrough-card`}
+    >
+      <header className="ui-WalkthroughCard--Title">{data.title}</header>
```
nicholaslyon-dt: Also I'm not a semantic expert, but I **_think_** the structure we want here is something like:

```html
<article>
  <header>
    <h2/h3> title -- depends on where it's being used
    <p> subtitle
  </header>
  <p> description
</article>
```

packages/data/walkthroughs/9.9.9.json
```diff
@@ -33,27 +35,11 @@
...
-      "sectionQuestions": [
-        "P18",
-        "P19",
-        "P20",
-        "P21",
-        "P22",
-        "P23",
-        "P24"
-      ]
+      "sectionQuestions": ["P18", "P19", "P20", "P21", "P22", "P23", "P24"]
```
nicholaslyon-dt, line 42: Did vscode auto do this for you or was this manual. My IDE formats it one on each line.

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,46 @@
+.ui-WalkthroughCard--CardContainer {
+  display: flex;
+  border-radius: var(--layout-border-radius-xlarge);
+  box-shadow: var(--surface-shadow-none), var(--surface-shadow-none);
```
nicholaslyon-dt: I'm seeing `box-shadow: var(--surface-shadow-medium);` in the designs. Also the definitions in `variables.css` already have both parts with a comma.

Screenshot 2024-06-17 at 11 15 07 AM

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,46 @@
...
+
+.ui-WalkthroughCard--Title {
+  color: var(--support-border-color-info);
+  font: var(--typography-bold-h5);
+  text-decoration: var(--walkthrough-card-title-text-decoration-hidden);
+  transition: text-decoration-color var(--transition-medium);
+  text-underline-offset: var(--walkthrough-card-title-text-underline-offset);
+}
+
+.ui-WalkthroughCard--CardContainer:hover {
```
nicholaslyon-dt: Can we use [data-hovered] to match the paradigms that react aria uses? Take a look at how button.css/link.css are handling it. We might want to also consider something for the [data-focus-visible] state as well.

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,46 @@
...
+  cursor: pointer;
+  box-shadow: var(--surface-shadow-medium), var(--surface-shadow-medium);
```
nicholaslyon-dt: Same as above, I'm seeing large for this state.

Screenshot 2024-06-17 at 11 15 19 AM

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,46 @@
...
+  margin: var(--layout-margin-medium);
+  gap: var(--layout-margin-medium);
+  border: var(--layout-border-width-none);
+  text-align: left;
+  transition: box-shadow var(--transition-medium);
+}
+
+.ui-WalkthroughCard--Title {
+  color: var(--support-border-color-info);
+  font: var(--typography-bold-h5);
```
nicholaslyon-dt: This looks like it needs to be h4 for the homepage cards and h5 for the result page cards based on the design.

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,46 @@
+.ui-WalkthroughCard--CardContainer {
+  display: flex;
+  border-radius: var(--layout-border-radius-xlarge);
+  box-shadow: var(--surface-shadow-none), var(--surface-shadow-none);
+  background-color: var(--surface-color-background-white);
+  max-width: var(--walkthrough-card-max-width);
+  flex-direction: column;
+  font-size: 14px;
```
nicholaslyon-dt: Do we need this font declaration here? I think all the content inside has it's own font declarations right? Same with the color line below this.

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,46 @@
...
+  color: var(--support-border-color-info);
+  padding: var(--layout-padding-large);
+  margin: var(--layout-margin-medium);
```
nicholaslyon-dt: I would assume in our use cases for these I'd want the container to some gap spacing to the cards. This makes a minimum spacing on these that can't be reduced by the parent easily.

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+  data,
+  href,
+}: WalkthroughCardProps) {
+  const router = useRouter();
+
+  const handleClick = () => {
+    if (href) router.push(href);
+  };
+
+  const canDisplayCard =
```
nicholaslyon-dt: I don't think we need this check based on all the types and properties being required. We'd get an error in useWalkthroughData.ts if any imported JSON didn't have the appropriate WalkthroughInfo.

apps/web/app/page.tsx
```diff
@@ -1,16 +1,24 @@
...
+        {Object.entries(WalkthroughJSONData).map(([id, { info }]) => {
+          return (
+            <WalkthroughCard
```
nicholaslyon-dt: These need a key property (ignore if you'd planned to cover that when building the actual page)

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,47 @@
+"use client";
+import { useRouter } from "next/navigation";
+
+import { Button } from "react-aria-components";
+
+import { WalkthroughInfo } from "../../../data/src/useWalkthroughData";
```
nicholaslyon-dt: This needs to import from the package.

`@repo/data/useWalkthroughData`

apps/web/app/page.tsx
```diff
@@ -1,16 +1,24 @@
 // 3rd party
 import { JSX } from "react";
 // repo
-import Link from "@repo/ui/link";
 import { URL_WALKTHROUGH_HREF } from "@repo/constants/src/urls";
-import ModalSide from "@repo/ui/modal-side";
+import WalkthroughCard from "@repo/ui/walkthrough-card";
+import { WalkthroughJSONData } from "../../../packages/data/src/useWalkthroughData";
```
nicholaslyon-dt: Needs to use the `@repo/data..` import

packages/data/src/useWalkthroughData.ts
```diff
@@ -75,7 +75,7 @@ export interface QuestionMultipleChoiceData extends QuestionBaseData {
 
 export const WalkthroughItemTypeMultiChoiceMultiple = "multiChoiceMultiple";
 export const isWalkthroughItemTypeMultiChoiceMultiple = (
-  walkthroughItemType: string,
+  walkthroughItemType: string
```
nicholaslyon-dt: My prettier puts this back in...

Can you run `npm run format` from the root? That _**should**_ format everything appropriately.

packages/data/src/useWalkthroughData.ts
```diff
@@ -122,15 +126,21 @@ export interface WalkthroughJSONType {
...
   "9.9.9": testCase999,
 };
 
-export default function useWalkthroughData({ id }: UseWalkthroughDataProps) {
-  // check if passed id is a key in WalkthroughJSONData
-  if (!Object.prototype.hasOwnProperty.call(WalkthroughJSONData, id)) {
+export default function useWalkthroughData({
+  id,
+}: UseWalkthroughDataProps): WalkthroughJSONType {
+  if (!id) {
```
nicholaslyon-dt: Since id is a required prop by our types, do we need this?

nicholaslyon-dt (changes requested):
Sorry for all the comments! It looks great overall.

packages/ui/src/walkthrough-card/WalkthroughCard.css
wardm89: Yep, that's what it transitions to on hover

apps/web/app/page.tsx
wardm89: Nice I didn't know @repo was configured to support that. This is why I mentioned adding absolute imports! haha

packages/data/walkthroughs/9.9.9.json
wardm89: Yes mine did that automatically when I saved.

nicholaslyon-dt: Alright. I updated my IDE to "wrap if long" hopefully it'll leave it as is from now on.

packages/ui/src/walkthrough-card/WalkthroughCard.css
wardm89: Since the card isn't a specific aria component like the button or link then no we can't use that...

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
wardm89: True

packages/data/src/useWalkthroughData.ts
wardm89: I think that's fine and the other comment about not needing the check as long as we are on the same page with the assumption that the walkthrough `json` is in fact correct and should bomb the application (in lower envs) if we somehow mess it up and pass incorrect data.

packages/ui/src/walkthrough-card/WalkthroughCard.css
nicholaslyon-dt: Ahh, previously it was using the Button component so we could have.

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+    if (href) router.push(href);
+  };
+
+  const canDisplayCard =
+    data && data.title && data.subtitle && data.description;
+
+  if (!canDisplayCard) return null;
+
+  return (
+    <a
```
nicholaslyon-dt: Sorry, I should have said this different the first time. This should use either the react-aria-components Link or the next/link to take advantage of client side routing. This way we can just add an href to it and it'll also be accessible. We don't need to handle any clicking or use the router here.

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
nicholaslyon-dt: Any reason I'm missing why we still have it then?

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useRouter } from "next/navigation";
+
+import { WalkthroughInfo } from "@repo/data/useWalkthroughData";
+
+import "./WalkthroughCard.css";
+
+export interface WalkthroughCardProps {
+  id: string;
+  data: WalkthroughInfo;
+  href?: string;
```
nicholaslyon-dt: I think this should be required based on our use cases. What would your thoughts be about making this be a property called `walkthroughId` and moving the pathing part with `URL_WALKTHROUGH_HREF` directly into the component. In theory the 2 use cases for this always point to a walkthrough. We could also create a helper function in `url.ts` to replace `URL_WALKTHROUGH_HREF` like we've done with some of the testids like so:

```TypeScript
export const GET_WALKTHROUGH_HREF = (id: string) => `/walkthrough/${id}/`;
```

packages/ui/src/walkthrough-card/WalkthroughCard.css
nicholaslyon-dt: I'm not seeing this changed. Thoughts on removing it?

packages/ui/src/walkthrough-card/WalkthroughCard.css
nicholaslyon-dt: I'm seeing a transition from medium -> large on hover. Are you seeing something different?

nicholaslyon-dt (changes requested):
Updated comments and callouts!

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
wardm89: I like the idea of moving it to the card actually. The helper function I don't think is necessary since we already have a constant for the base path for walkthroughs and can just do 
`href={`${URL_WALKTHROUGH_HREF}/${walkthroughId}`}`

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+export interface WalkthroughCardProps extends ReactAriaLinkProps {
+  id: string;
+  data: WalkthroughInfo;
+  walkthroughId: string;
+}
+
+export default function WalkthroughCard({
+  id,
+  data,
+  walkthroughId,
```
nicholaslyon-dt, line 22: Since we're extending the props, we should probably catch them here with `...props` and pass them along to ReactAriaLink.

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,45 @@
...
+  font: var(--typography-bold-h4);
+  text-decoration: var(--walkthrough-card-title-text-decoration-hidden);
+  transition: text-decoration-color var(--transition-medium);
+  text-underline-offset: var(--walkthrough-card-title-text-underline-offset);
+}
+
+.ui-WalkthroughCard--CardContainer[data-hovered] {
+  cursor: pointer;
+  box-shadow: var(--surface-shadow-large), var(--surface-shadow-large);
+  border: 1px solid var(--surface-color-border-default);
```
nicholaslyon-dt: Good call on the border on hover. Completely missed that before. Can we add it as a transparent border by default and then just change the border color when hovering? I think the animation looks better that way. Also update the transition declaration to include boder-color.

packages/ui/src/walkthrough-card/WalkthroughCard.css
```diff
@@ -0,0 +1,45 @@
...
+}
+
```
nicholaslyon-dt, line 35: Can we add a data-focus-visible like we do in button and link to make sure our focus outline is the same style as the rest of the site? That way we don't leave it up to the user to determine and get differing outlines?

My focus outline with green highlight color on MacOS
Screenshot 2024-06-18 at 11 09 04 AM

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+} from "react-aria-components";
+
+import { WalkthroughInfo } from "@repo/data/useWalkthroughData";
+import { URL_WALKTHROUGH_HREF } from "@repo/constants/src/urls";
+import "./WalkthroughCard.css";
+
+export interface WalkthroughCardProps extends ReactAriaLinkProps {
+  id: string;
+  data: WalkthroughInfo;
+  walkthroughId: string;
```
nicholaslyon-dt, line 16: Can we add the data-testid prop like we have in the other components to match the paradigm? Lauren has said this is helpful for selection on her side for the cypress tests as well.

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+export default function WalkthroughCard({
+  id,
+  data,
+  walkthroughId,
+}: WalkthroughCardProps) {
+  return (
+    <ReactAriaLink
+      className="ui-WalkthroughCard--CardContainer"
+      href={`${URL_WALKTHROUGH_HREF}/${walkthroughId}`}
+      aria-label={`${id}-walkthrough-card`}
```
nicholaslyon-dt: I like this as it doesn't read the whole card content aloud but I think we may need more context for the user. This is where I always struggle on determining enough context without too much. Something like:

`walkthrough ${id} - ${data.title}` ?

Would read out "walkthrough 9.9.9 [slight pause] Egress from Dwelling Units". Thoughts? Too much text/content?

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+    >
+      <article>
+        <header className="ui-WalkthroughCard--Title">{data.title}</header>
+        <span className="ui-WalkthroughCard--Subtitle">{data.subtitle}</span>
+        <section className="ui-WalkthroughCard--Description">
```
nicholaslyon-dt: I think this can just be a `<p>` element. We don't need a separate section for the description IMO.

Also from MDN for `<section>` element:

"Sections should always have a heading, with very few exceptions."

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

apps/web/app/page.tsx
```diff
@@ -1,15 +1,19 @@
...
 export default function Page(): JSX.Element {
   return (
     <div className="u-container">
-      <p>
-        <Link href={`${URL_WALKTHROUGH_HREF}/9.9.9/`}>Walkthrough 9.9.9</Link>
-      </p>
+      <div>
+        {Object.entries(WalkthroughJSONData).map(([id, { info }]) => {
+          return (
+            <WalkthroughCard key={id} id={id} data={info} walkthroughId={id} />
```
nicholaslyon-dt: Also now since `id` and `walkthroughId` are the same I'm not sure we need separate props for that.

packages/ui/src/walkthrough-card/WalkthroughCard.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+      ),
+    });
+
+    const linkElement = getByRole("link", {
+      name: `${testWalkthroughId}-walkthrough-card`,
+    });
+    expect(linkElement).toBeInTheDocument();
+  });
+
+  it("renders and navigates correctly on click", async () => {
```
nicholaslyon-dt: Needs renamed to reflect it renders appropriately with the href, we don't test navigation anymore (which I agree with, don't think we need to test default functionality)

nicholaslyon-dt (changes requested):
Sorrrrrrrrrry I found more annoying things.

nicholaslyon-dt (approved):
Love it. Looks fantastic!

PR #36, opened 2024-06-13
HOUSNAV-47: 9.9.9.1 ui questions

nicholaslyon-dt:
[HOUSNAV-47](https://hous-bssb.atlassian.net/browse/HOUSNAV-47)

## Overview & Purpose
Add the logic necessary for getting through the questions associated with 9.9.9.1

## Summary of Changes
Refactored currentItemId to be in the NavigationStore and to use a getter and setter for extra logic tied to when it's set
Added the ability for MultiChoiceMultiple data to be an array or an object
Added logic to dynamically show possible answers
Grouped more navigation state changes together
Added logic to handle questions that just set a variable
Marked some questions as not requiring an answer and added functionality for that
Added logic to set a default answer for some questions that are not required
Updated back logic to handle questions in history that just set a variable
Added type checking utilities
Added next navigation logic types and updated the base function to be recursive for and and or types
Added missing types in walkthrough data
Cleaned up walkthrough data errors and added new properties

## Side Effects
Could effect the intro questions, but I have not been skipping them to be able to test new questions

## Testing
Answer intro and questions in 9.9.9.1
Move back and forth changing answers and compare against [test case 1](https://docs.google.com/document/d/1VToXTJ10rf__5g771tb5KhCvihJxZPouY5XkYqEnLgk/edit)

### DIFF
--- a/apps/web/stores/AnswerStore.ts
+++ b/apps/web/stores/AnswerStore.ts
@@ -2,28 +2,116 @@
 import { makeAutoObservable } from "mobx";
 // local
 import { WalkthroughRootStore } from "./WalkthroughRootStore";
+import {
+  QuestionVariableData,
+  VariableToSetPropertyName,
+} from "@repo/data/useWalkthroughData";
+import {
+  answerValuesAreNotEqual,
+  getVariableItemValue,
+} from "../utils/logic/variableItem";
+import { isObject, isString } from "../utils/typeChecking";
 
-export type WalkthroughAnswerType = string | string[] | Record<string, string>;
-export type WalkthroughStoreAnswers = Record<string, WalkthroughAnswerType>;
+export type AnswerTypes = string | string[] | Record<string, string>;
+export type AnswerState = Record<string, AnswerTypes>;
+export type AnswerToCheckValueFn = (
+  answerToCheck: string,
+) => AnswerTypes | undefined;
 
 export const DEFAULT_ANSWER_VALUE_MULTI_CHOICE = undefined;
-export const DEFAULT_ANSWER_VALUE_MULTI_CHOICE_MULTI = [];
+export const DEFAULT_ANSWER_VALUE_MULTI_CHOICE_MULTI: string[] = [];
 
 export class AnswerStore {
   rootStore: WalkthroughRootStore;
 
-  answers: WalkthroughStoreAnswers = {};
+  answers: AnswerState = {};
 
   constructor(rootStore: WalkthroughRootStore) {
     makeAutoObservable(this);
     this.rootStore = rootStore;
   }
 
-  setAnswerValue = (value: WalkthroughAnswerType, questionId: string) => {
+  handleVariableItem = (
+    currentItem: QuestionVariableData,
+    currentId: string,
+  ) => {
+    // get state references
+    const {
+      navigationStore: { updateNavigationState, addItemIdToHistory },
+    } = this.rootStore;
+
+    // get new variable value
+    const newItemValue = getVariableItemValue(
+      currentItem[VariableToSetPropertyName],
+      this.getAnswerToCheckValue,
+    );
+
+    // get current variable value
+    const currentItemValue = this.answers[currentId];
+
+    // compare with current variable value to see if it's changed
+    // if changed, update variable value and erase future state
+    if (
+      !currentItemValue ||
+      (currentItemValue &&
+        answerValuesAreNotEqual(currentItemValue, newItemValue))
+    ) {
+      // erase future state
+      this.eraseAnswersAfterCurrentItem();
+
+      // set answer value
+      this.answers[currentId] = newItemValue;
+    }
+
+    // add variable item to history
+    addItemIdToHistory(currentId);
+
+    // update navigation state again
+    updateNavigationState(currentItem.nextNavigationLogic);
+  };
+
+  // NOTE: This is only called onChange in Question component types
+  setAnswerValueOnChange = (value: AnswerTypes, questionId: string) => {
+    // erase future state
+    this.eraseAnswersAfterCurrentItem();
+
+    // set answer value
+    this.answers[questionId] = value;
+  };
+
+  setDefaultAnswerValue = () => {
+    // get references
+    const {
+      navigationStore: { currentItemId },
+      currentQuestionAsMultipleChoiceMultiple,
+      currentPossibleAnswersFromMultipleChoiceMultiple,
+    } = this.rootStore;
+
+    // set default value for new item if it's a multiple choice multiple question and isNotRequired
+    if (
+      currentQuestionAsMultipleChoiceMultiple &&
+      currentQuestionAsMultipleChoiceMultiple.isNotRequired
+    ) {
+      if (currentQuestionAsMultipleChoiceMultiple.storeAnswerAsObject) {
+        // store answer as object
+        this.answers[currentItemId] =
+          currentPossibleAnswersFromMultipleChoiceMultiple.reduce<
+            Record<string, string>
+          >((acc, curr) => {
+            acc[curr.answerValue] = "false";
+            return acc;
+          }, {});
+      } else {
+        // store answer as array
+        this.answers[currentItemId] = DEFAULT_ANSWER_VALUE_MULTI_CHOICE_MULTI;
+      }
+    }
+  };
+
+  eraseAnswersAfterCurrentItem = () => {
     // get state of other store variables
     const {
-      navigationStore: { questionHistory },
-      currentItemId,
+      navigationStore: { questionHistory, currentItemId },
     } = this.rootStore;
 
     // if current question is NOT last one in questionHistory, remove all questions after current question and their answers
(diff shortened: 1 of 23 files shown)


apps/web/utils/logic/showAnswer.ts
```diff
@@ -0,0 +1,77 @@
...
+// local
+import { AnswerTypes, AnswerToCheckValueFn } from "../../stores/AnswerStore";
+import { isArray } from "../typeChecking";
+
+export const getPossibleAnswers = (
+  possibleAnswers: PossibleAnswer[],
+  getAnswerToCheckValue: AnswerToCheckValueFn,
+) => {
+  return possibleAnswers.filter((possibleAnswer) => {
+    if (!possibleAnswer.showAnswerIf) return true;
```
jbook24: Could we do something like this? 
`if (possibleAnswer.showAnswerIf == null || possibleAnswer.showAnswerIf === true) return true;
`

apps/web/utils/logic/showAnswer.ts
```diff
@@ -0,0 +1,77 @@
...
+) => {
+  // setup showAnswer variable
+  let showAnswer = false;
+
+  // check if answerToCheck is greater than answerValue
+  if (answerToCheck > answerValue) {
+    showAnswer = true;
+  }
+
+  return showAnswer;
```
jbook24: Could we just ? 

```
// Return true if answerToCheck is greater than answerValue, otherwise false
return answerToCheck > answerValue;
```

apps/web/utils/logic/variableItem.ts
```diff
@@ -0,0 +1,110 @@
...
+import { isArray, isObject, isString } from "../typeChecking";
+// NOTE: this feels weird, but it makes sure this module is something we can spy on for tests
+import * as ThisModule from "./variableItem";
+
+export const getVariableItemValue = (
+  variableToSet: VariableToSet,
+  getAnswerToCheckValue: AnswerToCheckValueFn,
+): AnswerTypes => {
+  // check variableType
+  switch (variableToSet.variableType) {
```
jbook24: Given we only have one case in our switch could we just? 
```

if (variableToSet.variableType === VariableToSetType.Object) {
  return ThisModule.getVariableItemValueObject(variableToSet, getAnswerToCheckValue);
}

```

apps/web/utils/logic/showAnswer.ts
nicholaslyon-dt: Not quite because of a quirk with TypeScript and JSON. the `showAnswerIf` property can be `boolean | ShowAnswerIf[] | undefined`. But really, it can never be false. I just can't set it to `true | ShowAnswerIf[] | undefined` because of the type difference issues between TypeScript and JSON. I will update this to absorb the check below it though. LMK what you think.

apps/web/utils/logic/showAnswer.ts
nicholaslyon-dt: Yes! Good call.

apps/web/utils/logic/variableItem.ts
nicholaslyon-dt: Yeah, I had left this as is because there was a potential for another type, but at the moment there is not!

wardm89:
Screenshot 2024-06-14 at 2 21 53 PM
I just got this error. So far, I'm trying to replicate it but idr my answers I did to get it :) But I do know it was after clicking `2 stories`.

Any thoughts on this?
I'm assuming it came through undefined.

nicholaslyon-dt:
> Screenshot 2024-06-14 at 2 21 53 PM I just got this error. So far, I'm trying to replicate it but idr my answers I did to get it :) But I do know it was after clicking `2 stories`.
> 
> Any thoughts on this? I'm assuming it came through undefined.

Hrm I'm not sure. Possible Invalid Answers should only be called in 3 cases and they all happen in the intro questions. I've been running through the questions back and forth and can't get that to happen.

apps/web/utils/logic/nextNavigation.ts
```diff
@@ -38,106 +81,249 @@ export const getNextNavigationId = (
...
+      isNotEqual = true;
+    }
+  } else {
+    // if answerToCheck is not a string or undefined, throw an error
+    throw new Error(
+      `nextLogicTypeNotEqual: answerToCheck must be a string or undefined, got ${typeof answerToCheck}`,
+    );
+  }
+
+  return isNotEqual;
```
wardm89:
```suggestion
  if (isString(answerToCheck)) {
    return answerToCheck !== answerValue;
  }

  if (answerToCheck === undefined) {
    return answerValue !== "undefined";
  }
  // if answerToCheck is not a string or undefined, throw an error
  throw new Error(
    `nextLogicTypeNotEqual: answerToCheck must be a string or undefined, got ${typeof answerToCheck}`
  );
```

apps/web/utils/logic/nextNavigation.ts
```diff
@@ -38,106 +81,249 @@ export const getNextNavigationId = (
...
+  } else if (answerToCheck === undefined) {
+    if (answerValue === "undefined") {
+      isEqual = true;
+    }
+  } else {
+    // if answerToCheck is not an array, string, or undefined, throw an error
+    throw new Error(
+      `nextLogicTypeEqual: answerToCheck must be a string, array, or undefined, got ${typeof answerToCheck}`,
+    );
   }
```
wardm89: Very similar recommendation to the other. We don't actually need all the else if statements and can simplify this.

apps/web/utils/logic/showAnswer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      // if showAnswerLogicType is not recognized, throw an error
+      throw new Error(
+        `getPossibleAnswers: showAnswerLogicType not recognized: ${showAnswerIf.showAnswerLogicType}`,
+      );
+    });
+  });
+};
+
+export const showAnswerTypeEquals = (
```
wardm89, line 49: This can be shortened to remove the if else statements as well.

if(isArray)return...
if(isString)return...
throw error

apps/web/utils/logic/variableItem.ts
```diff
@@ -0,0 +1,110 @@
...
+      variableToSet,
+      getAnswerToCheckValue,
+    );
+  }
+
+  // throw error if variableType is not valid
+  throw new Error("Invalid variableType");
+};
+
+export const getVariableItemValueObject = (
```
wardm89, line 27: I feel like this could be shortened a lot by deconstructing the variables and then not needing `valueHasBeenSet` since we can just `break` the loop after each `newObjectVariable[key] =` line.

apps/web/utils/logic/nextNavigation.ts
```diff
@@ -1,31 +1,74 @@
...
+    // if no nextLogicType is found, throw an error
+    throw new Error(
+      `nextLogicType ${navigationLogicItem.nextLogicType} is not supported for valuesToCheck property.`,
+    );
+  } else if (navigationLogicItem.answerToCheck) {
     // get answer to check
-    const answerToCheck = answerState[navigationLogicItem.answerToCheck || ""];
+    const answerToCheck = getAnswerToCheckValue(
+      navigationLogicItem.answerToCheck,
+    );
```
wardm89: What is the difference between this `answerToCheck` and the one that could be deconstructed from `navigationLogicItem`?

apps/web/stores/AnswerStore.ts
```diff
@@ -2,28 +2,116 @@
...
+    if (
+      !currentItemValue ||
+      (currentItemValue &&
+        answerValuesAreNotEqual(currentItemValue, newItemValue))
+    ) {
+      // erase future state
+      this.eraseAnswersAfterCurrentItem();
+
+      // set answer value
+      this.answers[currentId] = newItemValue;
```
wardm89: This goes for a lot of these comments. I feel they are redundant. I think you've done a good job naming functions/methods as well so things like this below is also redundant.
```
// add variable item to history
 addItemIdToHistory(currentId);
```

wardm89 (changes requested):
I think with some refactoring to deconstruct variables so lines don't wrap unnecessarily, simplifying some if/else logic and removing comment's it'll make this a lot more readable. 

As far as my "bug" I had before. I can't replicate it but with the new error handling on navigation that will enable us to replicate it if it happens again.

wardm89:
One last comment. With all the errors being thrown, they aren't being handled anywhere currently or are they?

apps/web/utils/logic/nextNavigation.ts
nicholaslyon-dt: That's the value. Refactored across the board for clarity.

apps/web/stores/AnswerStore.ts
nicholaslyon-dt: Refactored across the board! LMK what you think.

nicholaslyon-dt:
> One last comment. With all the errors being thrown, they aren't being handled anywhere currently or are they?

Good call. Most were caught by the handling we added for the forward navigation state, but there was one case that still needed caught. I added a function for handling and caught it in the other spot as well.

wardm89 (approved):

jbook24 (approved):

PR #39, opened 2024-06-17
HOUSENAV-23: Intro UI Cypress Tests

laurenrevere:
[HOUSNAV-23](https://hous-bssb.atlassian.net/browse/HOUSNAV-23)

## Overview & Purpose
Add walkthrough tests for the intro UI.

## Summary of Changes

- Added walkthroughs for most cases in the intro UI that kick the user out to a a result.
- Update the UI tests to deal with checkboxes.
- Move the results data into its own object.
- Add custom command for grabbing inputs specifically.

## Notes & Resources
Defined terms are causing some problems for the UI tests. They are easy for cypress to accidentally click one of them instead of the actual button. When we run the tests focusing on the checkbox instead of the center of the buttons, it works.

### DIFF
--- a/apps/web/cypress/e2e/walkthroughs.cy.ts
+++ b/apps/web/cypress/e2e/walkthroughs.cy.ts
@@ -19,10 +19,12 @@ describe("walkthrough 1", () => {
         if (step.type === "radio") {
           cy.getByTestID(GET_TESTID_RADIO(step.question, step.answer)).click();
         } else if (step.type === "checkbox") {
-          // TODO - update the application to not have duplicate test ids on checkboxes
-          cy.getByTestID(GET_TESTID_CHECKBOX(step.question, step.answer))
-            .last()
-            .click({ force: true });
+          step.answer.split(",").forEach((answer) => {
+            // TODO - update the application to not have duplicate test ids on checkboxes
+            cy.getByTestID(GET_TESTID_CHECKBOX(step.question, answer))
+              .last()
+              .click({ force: true });
+          });
         }
         cy.getByGeneralTestID(TESTID_WALKTHROUGH_FOOTER_NEXT).click();
       });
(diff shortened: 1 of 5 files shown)


apps/web/cypress/e2e/walkthroughs.cy.ts
```diff
@@ -11,17 +15,45 @@ describe("walkthrough 1", () => {
...
     it(walkthrough.title, () => {
       walkthrough.steps.forEach((step) => {
-        // TODO - use test ids
         // select and submit an answer for the given question
-        cy.getByTestID(`${step.type}-${step.question}-${step.answer}`).click();
+        if (step.type === "radio") {
+          cy.getByTestID(GET_TESTID_RADIO(step.question, step.answer)).click();
+        } else if (step.type === "checkbox") {
+          step.answer.split(",").forEach((answer) => {
+            // TODO - update the application to not have duplicate test ids on checkboxes
```
nicholaslyon-dt: This has to do with the way aria react adds that property down the chain of elements. I'm not sure we can fix it easily from our end. What if we introduced a new helper to specifically select an input element by test id?

```javascript
Cypress.Commands.add("getInputByTestID", (id, options) => {
  return cy.get(`input[data-testid=${id}]`, options);
});
```

nicholaslyon-dt (approved):

apps/web/cypress/e2e/walkthroughs.cy.ts
laurenrevere: I could have it deliberately get the "last" element with that ID, but still feels a little clunky. Thoughts?

laurenrevere: Good suggestion!

nicholaslyon-dt (approved):
Looks great!

PR #41, opened 2024-06-18
HOUSENAV-21: Walkthrough Card UI Tests

laurenrevere:
[HOUSNAV-21](https://hous-bssb.atlassian.net/browse/HOUSNAV-21)

## Overview & Purpose
Added in a cypress test for the home page. 

## Summary of Changes
- Added test to confirm navigation to walkthrough from home page
- Rename cypress test files to separate them by page/walkthrough
- Added accessibility scan on home page
- Update walkthrough card test id so that it is able to be unique per card
- Fixed [this build failure](https://github.com/bcgov/House-Innovations-HOUSNAV/actions/runs/9571303244/job/26394225313)

## Side Effects
All the cypress commands now handle a more diverse set of test ids. I made some changes to walkthrough card unit tests to accommodate the change in test id.

By casting `walkthrough.result` to a key of results, we are skipping the opportunity to add type checking for the walkthrough objects. I believe that this is okay in the testing context because the test will promptly fail if the following cypress command is passed an "undefined" or incorrect value.

## Notes & Resources
The accessibility scan will have to be enabled once the [home page structure is completed](https://hous-bssb.atlassian.net/browse/HOUSNAV-72). Right now accessibility would fail because the HTML structure is not complete for the page.

## Checklist
- [ ] I have written or updated vitests for this work
- [ ] I have reviewed the [accessibility guidelines](https://digital.gov.bc.ca/wcag/home/) relevant to this work
- [ ] I have reviewed this work with at least one screen reader (when applicable)
- [ ] I have added/updated any related documentation


packages/ui/src/walkthrough-card/WalkthroughCard.tsx
```diff
@@ -20,7 +20,7 @@ export interface WalkthroughCardProps extends ReactAriaLinkProps {
 export default function WalkthroughCard({
   data,
   walkthroughId,
-  "data-testid": testid = TESTID_WALTHROUGH_CARD,
+  "data-testid": testid = GET_TESTID_WALKTHROUGH_CARD(walkthroughId),
```
nicholaslyon-dt: Looking at this more now I think we want to refactor to this:

// get passed unique test id or default to nothing
`"data-testid": testid = "",`

// make testid unique if an override is passed, otherwise default to the walkthroughId
`data-testid={GET_TESTID_WALKTHROUGH_CARD(testid || walkthroughId)}`

As this is set up now, if a testid is passed it will lose the `walkthrough-card-` prefix. I would think we'd want all testids for the walkthrough cards to contain the prefix.

nicholaslyon-dt (changes requested):
Only one small tweak.

packages/ui/src/walkthrough-card/WalkthroughCard.tsx
laurenrevere: Updated!

nicholaslyon-dt (approved):

PR #42, opened 2024-06-20
HOUSNAV-48: UI Questions 9.9.9.2

nicholaslyon-dt:
[HOUSNAV-48](https://hous-bssb.atlassian.net/browse/HOUSNAV-48)

## Overview & Purpose
Updates needed to facilitate handling all the questions in section 9.9.9.2 of Walkthrough 9.9.9.

## Summary of Changes
Created PDFDownloadLink component to handle links in content that point the Building Code PDF
Updated string parsing to handle new PDFDownloadLink component
Added GreaterThan logic in NextNavigation

## Testing
Navigate to section 9.9.9.2 (starts with question P11). 

## Notes & Resources
- You can jump to 9.9.9.1 and start there by updated the `startingSectionId` in 9.9.9.json to `9991`.
- Question P15 contains new PDFDownloadLink component
- [Walkthrough 9.9.9 Document](https://docs.google.com/document/d/1VToXTJ10rf__5g771tb5KhCvihJxZPouY5XkYqEnLgk/edit)

## Checklist

### DIFF
--- a/apps/web/utils/string.tsx
+++ b/apps/web/utils/string.tsx
@@ -55,41 +55,33 @@ export const parseStringToComponents = (
     replace: (domNode: DOMNode) => {
       if (domNode instanceof Element && domNode.attribs) {
         switch (domNode.name) {
-          case definedTermName:
-            {
-              const DefinedTermComponent = customComponents[definedTermName];
-              const props = attributesToProps(domNode.attribs);
+          case definedTermName: {
+            const DefinedTermComponent = customComponents[definedTermName];
+            const props = attributesToProps(domNode.attribs);
 
-              if (DefinedTermComponent) {
-                return (
-                  <DefinedTermComponent
-                    {...(props as unknown as DefinedTermProps)}
-                    key={domNode.attribs.key}
-                  >
-                    {domToReact(domNode.children as DOMNode[], options)}
-                  </DefinedTermComponent>
-                );
-              }
-            }
-            break;
-          case pdfDownloadLinkName:
-            {
-              const PDFDownloadLinkComponent =
-                customComponents[pdfDownloadLinkName];
-              const props = attributesToProps(domNode.attribs);
+            return (
+              <DefinedTermComponent
+                {...(props as unknown as DefinedTermProps)}
+                key={domNode.attribs.key}
+              >
+                {domToReact(domNode.children as DOMNode[], options)}
+              </DefinedTermComponent>
+            );
+          }
+          case pdfDownloadLinkName: {
+            const PDFDownloadLinkComponent =
+              customComponents[pdfDownloadLinkName];
+            const props = attributesToProps(domNode.attribs);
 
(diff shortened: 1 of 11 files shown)


apps/web/utils/string.tsx
```diff
@@ -53,6 +72,24 @@ export const parseStringToComponents = (
               }
             }
             break;
+          case pdfDownloadLinkName:
+            {
+              const PDFDownloadLinkComponent =
+                customComponents[pdfDownloadLinkName];
+              const props = attributesToProps(domNode.attribs);
+
+              if (PDFDownloadLinkComponent) {
```
jbook24: What scenario would this be false?

nicholaslyon-dt: Good call! Not needed anymore since I refactored the type for the `customComponents` object. Removed!

jbook24:
Testing 
- With P(11) being greater than 2 

I am not able to navigate to R(2) as in the screenshot below 
image

jbook24:
Everything looks good from my side as an initial pass that we can iterate on later. 

One point I would like to raise is the warning that is appearing when navigating from question to question. 

**Attached is a snippet** 
image

**Potential Implications** 
When a component changes from uncontrolled to controlled, it means that at some point, the component's state was managed internally (uncontrolled) and then switched to being managed externally via props (controlled). This inconsistency can lead to unpredictable behavior and bugs, as the component might not handle state updates correctly.

**Potential Solution** 
Can we make a position of either "Always Controlled" or "Always Uncontrolled" ?

wardm89:
> Everything looks good from my side as an initial pass that we can iterate on later.
> 
> One point I would like to raise is the warning that is appearing when navigating from question to question.
> 
> **Attached is a snippet** image
> 
> **Potential Implications** When a component changes from uncontrolled to controlled, it means that at some point, the component's state was managed internally (uncontrolled) and then switched to being managed externally via props (controlled). This inconsistency can lead to unpredictable behavior and bugs, as the component might not handle state updates correctly.
> 
> **Potential Solution** Can we make a position of either "Always Controlled" or "Always Uncontrolled" ?

I will work on updating this. It seems to actually be because the radio buttons are set to an undefined state when the question is loaded and the once the selection is made it's state is then controlled. I'm not sure if it's related to Mobx or not but I'll look into it.

wardm89 (commented):
LGTM other than the one issue I'll look into and also `type CustomComponentProps` I'm thinking we will want to rethink how we are handling these types so it's cleaner and more sustainable over time. However, I think it's fine for now and works so no need to change anything yet. It can be refactored when the complexity reaches the proverbial "tipping point".

wardm89 (approved):

wardm89:
Merging since I discussed this with Joshua that it can be merged. 

As a note my change was removing the "defaultValue" for the radio button group since the same functionality is accomplished with "value" when it's passed. Default values are for uncontrolled components so we shouldn't use them in conjunction like this.

PR #46, opened 2024-06-25
HOUSNAV-39: Glossary Terms DT + Modal refactor

wardm89:
[HOUSNAV-39](https://hous-bssb.atlassian.net/browse/HOUSNAV-39)

## Overview & Purpose

Add glossary term data transformation enabling correct tooltips and adds real data to modal popout.

## Summary of Changes

- Update references to open building code model
    - Updated/added new building code data to match potential format
- Added new cases for `parseStringToComponents` 
- Added `terms.json` to house all glossary terms
    - Moved to `json` directory instead of just `walkthroughs`   
- Refactored Modal so that the building code and glossary terms data is formatted separately. (Easier to debug this way) 
## Side Effects

- Tooltips had to be modified to account for the `major occupancy` terms and `groups`. So, tooltip content needed an override. If there is a defined term that has the incorrect tooltip then:
    - Check the `terms.json` and it's "clean definition" 
    - If that is correct then check what question/answer is using the defined term and ensure there isn't a `override-tooltip` set. If there is, ensure it's correct.

     
## Testing

- Test by opening glossary modal. Ensure tooltips and clicking on terms are correct.
- UI walkthroughs should contain terms
    -  **If walkthrough is missing a term then note the questions number and what the term is** 
    - I updated some of the defined terms in 9.9.9. If a term is a variation (ie. occupancy/occupancies) and the term does not have the correct tooltip or open the modal and scroll to it then this can easily be fixed by adding "override-term" to the attribute in the 9.9.9 json 
        - This was done so that we can quickly specify and override terms and it ensures the application is fast because it doesn't have to "loop" variations to see if one is correct. 
 - Note: Groups should display the definition or word for that group. However, when clicking on a group as a defined term it should open the glossary modal at the `major occupancy` section. 

## Checklist
- [ ] I have reviewed the [accessibility guidelines](https://digital.gov.bc.ca/wcag/home/) relevant to this work
    - N/A due to no new functionality 
- [ ] I have reviewed this work with at least one screen reader (when applicable)
    - Screen reader enhacements (https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3/backlog?selectedIssue=HOUSNAV-89) 

### DIFF
--- a/apps/web/utils/string.tsx
+++ b/apps/web/utils/string.tsx
@@ -13,11 +13,14 @@ import Tooltip from "@repo/ui/tooltip";
 import DefinedTerm, {
   DefinedTermProps,
 } from "../components/defined-term/DefinedTerm";
-import PDFDownloadLink, {
-  PDFDownloadLinkProps,
-} from "../components/pdf-download-link/PDFDownloadLink";
-import { GLOSSARY_TERMS } from "../tests/mockData";
-import { TooltipGlossaryData } from "@repo/data/useGlossaryData";
+import Button from "@repo/ui/button";
+import Tooltip from "@repo/ui/tooltip";
+import {
+  BuildingCodeJSONData,
+  ModalSideDataEnum,
+  TooltipGlossaryData,
+} from "@repo/data/useGlossaryData";
+import ModalSide from "@repo/ui/modal-side";
 
 // Define custom components for html-react-parser
 const definedTermName = "defined-term";
@@ -57,10 +60,13 @@ export const parseStringToComponents = (
   const options = {
     replace: (domNode: DOMNode) => {
       if (domNode instanceof Element && domNode.attribs) {
-        const overrideTerm = domNode.attribs["override-term"];
-        const term =
+        const term = (
           domNode.attribs["override-term"] ??
-          (domToReact(domNode.children as DOMNode[]) as string);
+          (domToReact(domNode.children as DOMNode[]) as string)
+        ).toLocaleLowerCase();
+        const tooltipTerm = (
+          domNode.attribs["override-tooltip"] ?? term
+        ).toLocaleLowerCase();
         switch (domNode.name) {
           case definedTermName: {
             const DefinedTermComponent = customComponents[definedTermName];
@@ -70,8 +76,8 @@ export const parseStringToComponents = (
                   <DefinedTermComponent
                     {...(props as unknown as DefinedTermProps)}
                     key={domNode.attribs.key}
-                    overrideTerm={overrideTerm}
-                    term={domToReact(domNode.children as DOMNode[]) as string}
+                    term={term}
+                    overrideTooltip={tooltipTerm}
                   >
                     {domToReact(domNode.children as DOMNode[], options)}
                   </DefinedTermComponent>
@@ -94,7 +100,7 @@ export const parseStringToComponents = (
           case definedTermModal: {
             return (
               <Tooltip
-                tooltipContent={TooltipGlossaryData.get(term)}
+                tooltipContent={TooltipGlossaryData.get(tooltipTerm)}
                 triggerContent={
                   <Button
                     variant="glossary"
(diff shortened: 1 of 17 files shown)


jbook24:
I think this looks good! Is it possible to implement a smooth scroll when clicking different definitions?

apps/web/utils/string.tsx
```diff
@@ -82,27 +98,40 @@ export const parseStringToComponents = (
...
+              </span>
+            );
+
+          case buildingCode:
+            return (
+              <ModalSide
+                type={ModalSideDataEnum.BUILDING_CODE}
+                triggerContent={<Button variant="code">{term}</Button>}
+                modalData={BuildingCodeJSONData}
+                scrollTo={"9.9.9.1"}
```
jbook24, line 132: I see 9.9.9.1 is hardcoded - Will this implementation work for additional workflows?

packages/data/json/buildingCode.json
```diff
@@ -0,0 +1,150 @@
+[
+  {
+    "section": "9.9.9.1",
+    "header": "Egress from Dwelling Units",
+    "content": {
+      "clauses": [
+        {
+          "description": "Except as provided in <tag>Sentences (2) and (3), every dwelling unit containing more than 1 storey shall have exits or egress doors located so that it shall not be necessary to travel up or down more than 1 storey to reach a level served by...",
+          "subsections": [
```
jbook24, line 9: Thoughts on the children of clauses would have "subclauses" instead of "subsections" ?

packages/data/src/useWalkthroughTestData.ts
```diff
@@ -29,6 +29,23 @@ export function getMultiChoiceQuestion() {
   };
 }
 
+export function getQuestion(questionId: string) {
+  // find the question for the given questionId
+  const questions: {
+    [key: string]: QuestionDisplayData | QuestionVariableData;
+  } = testCase999.questions;
```
jbook24, line 36: Will this only work for Workflow 1?

packages/ui/src/modal-side/ModalSide.tsx
```diff
@@ -10,31 +10,191 @@ import {
...
+        <li key={index}>
+          {subsection.description}
+          {subsection.subClauses && renderSubClauses(subsection.subClauses)}
+        </li>
+      ))}
+    </ol>
+  );
+}
+
+// TODO: (ANY) Move to GlossaryContent component
```
jbook24, line 72: When do we anticipate this being addressed?

packages/ui/src/modal-side/ModalSide.tsx
```diff
@@ -10,31 +10,191 @@ import {
...
+              ref={(el) => {
+                sectionRefs.current[data.section] = el;
+              }}
+              className={`ui-ModalSide--Section ${
+                highlightedSection === data.section
+                  ? "ui-ModalSide--SectionHighlighted"
+                  : ""
+              }`}
+            >
+              <article className="ui-ModalSide--SectionGlossaryContentLine">
```
jbook24, line 108: Love the Html5 semantics!

packages/ui/src/modal-side/ModalSide.tsx
```diff
@@ -75,39 +235,22 @@ export default function ModalSide({
...
-                        <p className="ui-ModalSide--SectionContent">
-                          {parseStringToComponents(
-                            section.content,
-                            setFocusSection,
-                          )}
-                        </p>
-                      </article>
-                    </section>
-                  ))}
+                  {type === ModalSideDataEnum.GLOSSARY && (
```
jbook24, line 238: Love the implementation here!

packages/data/src/useGlossaryData.ts
```diff
@@ -0,0 +1,69 @@
...
+  content: ArticleContentType | GlossaryContentType;
+};
+
+export type BuildingGlossaryJSONType = {
+  [term: string]: GlossaryContentType;
+};
+
+const GlossaryJSONData = glossary as unknown as BuildingGlossaryJSONType;
+
+// TODO: (ANY) Move to useBuildingCodeData
```
jbook24, line 44: When will this be addressed?

jbook24 (approved):

apps/web/utils/string.tsx
wardm89: This will not be hardcoded once the Building code stuff is done. It would "scroll to" the correct building code section

packages/data/json/buildingCode.json
wardm89: naming is updated in BC ticket

packages/data/src/useWalkthroughTestData.ts
wardm89: Technically yes but since that file is only used essentially as a "test" helper to test the handling of various types of questions it's ok since it doesn't actually matter what workflow it is here.

packages/ui/src/modal-side/ModalSide.tsx
wardm89: It's not mandatory but I think it would be good to add at some point maybe in the polish tickets once the modal is done being changed. I had to stop writing code at some point since my ticket was only about Glossary DT hahaa

packages/data/src/useGlossaryData.ts
wardm89: Same as above.

PR #54, opened 2024-07-03
HOUSNAV-38: 9.9.9.1-3 DT

wardm89:
[HOUSNAV-38](https://hous-bssb.atlassian.net/browse/HOUSNAV-38)

## Overview & Purpose

Add 9.9.9.1-3 Building code data and update modal to account for new data structure

## Summary of Changes

- Add 9.9.9.1-3 Building code data
- Update Building Code JSON structure to handle everything from `Parts` -> `subclauses`
- Break up Glossary and Building Code content into separate component files making it easier to maintain
- Updates building code component to have specific render functions at each level to reduce complexity with all the nesting
- Updates some Modal CSS (Still not 100% polished)

## Side Effects

- Updates terms to reference the pdf download instead of the building code
- Added a *partial* 9.10.14.4 to test the image use case

## Testing

https://github.com/bcgov/House-Innovations-HOUSNAV/assets/7059124/7e90096b-0f8c-4847-81f6-4d7b50cdc1eb

## Checklist
    - Screen reader still not working right for the Modal (Fix in Polish)


packages/data/json/buildingCode.json
```diff
@@ -1,150 +1,194 @@
...
+                    "clauses": [
+                      {
+                        "numberReference": "9.9.9.3.(3)(a)",
+                        "description": "a balcony, or"
+                      },
+                      {
+                        "numberReference": "9.9.9.3.(3)(b)",
+                        "description": " an openable window conforming to sentences <building-code-modal reference='9.9.9.1.(2)(a)'>9.9.9.1.(2)(a)</building-code-modal> and <building-code-modal reference='9.9.9.1.(2)(b)'>(b)</building-code-modal>."
```
laurenrevere: I believe in the building code this says "Clauses" instead of "sentences" in this spot. (I'm not quire sure why though..)

laurenrevere:
These three sections look really good!

packages/data/json/9.10.14.json
nicholaslyon-dt, line 1: This happened while I was out and I meant to comment on it earlier. Any reason why the walkthroughs JSON was moved into here and why all the JSON is now lumped in 1 folder? Seems confusing that 9.9.9 is in here named as such. I would initially question whether it was building code JSON or Walkthrough data JSON.

packages/data/json/9.9.9.json
```diff
@@ -457,7 +457,7 @@
       "questionText": "Is this location a <defined-term>dwelling unit</defined-term>?",
       "questionCodeReference": {
         "displayString": "Vol 2, Div B, Article 9.9.9.1.",
-        "codeNumber": "V2-DB-9.9.9.1"
+        "codeNumber": "9.9.9.1"
```
nicholaslyon-dt: I know this is just in theory, but I had the volume and division in the code number to future proof the lookup for when there may be duplicate part number. Looking at the table of contents again we really only need division since Volume 2 only has Part 9, which Volume 1 doesn't have. Not sure it matters at this point but we could leave the division part in for future purposes.

Screenshot 2024-07-08 at 1 07 52 PM

packages/data/json/buildingCode.json
nicholaslyon-dt, line 1: Is the intent here that ALL the building code content goes here and we only have 1 building code modal? The design only shows the building code that applies to the walkthrough the user is currently viewing. Meaning they wouldn't be able to scroll down and see 9.10.14 while looking at the building code modal in walkthrough 9.9.9. I was assuming the modal would take a parameter to show only the building code sections that it needed to for the current walkthrough. Thoughts?

packages/ui/src/modal-side/ModalSide.css
```diff
@@ -86,33 +78,38 @@
...
+.ui-ModalSide--SectionNumberLarge {
+  font-size: var(--typography-font-size-h3);
 }
 
-.ui-ModalSide--SectionNumber {
-  grid-column: span 1;
-  text-align: center;
-  font-size: var(--typography-font-size-h5);
+.ui-ModalSide--SectionHeader,
+.ui-ModalSide--SectionHeaderLarge {
```
nicholaslyon-dt, line 99: I think we need a little style work on the mobile version. Some things I'm noticing:

1. Spacing with outside edge is too large.
2. Font size and line height of bold headers is too large
3. In the screenshot below you can see the Article 9.9.9.1 header is actually indented from the Subsection 9.9.9 header.

Screenshot 2024-07-08 at 1 19 01 PM

nicholaslyon-dt: Some of the same things apply to desktop as well but I also think desktop _**looks**_ fine so I'm cool with leaving this as is.

Screenshot 2024-07-08 at 1 25 06 PM

packages/ui/src/modal-glossary-content/ModalGlossaryContent.tsx
nicholaslyon-dt, line 1: There are a lot of overlapping `<section>` elements created. I think this needs a refactor for semantics. I think I could make the argument that the entire modal content itself is an `<article>` as it is a 'self-contained composition... which is intended to be independently distributable' [MDN Article Page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article). And in that case the Subsection and Article headers should be an H2 and H3 respectively. I could also see an argument for each Article being it's own `<article>` (haaa 😄). Thoughts?

Screenshot 2024-07-08 at 1 27 27 PM

nicholaslyon-dt (commented):
Overall this is great! I have a couple of comments (any of which could be moved into a cleanup ticket).

packages/data/json/buildingCode.json
wardm89: That's the current intent as I didn't see that as a requirement to only display the BC data for the section selected. I did think about that but since this is a POC the added complexity didn't make sense to me since there are several places in the building code that reference sections outside of the current section. That's currently being handled with the pdf link. To future proof it we'd need to think through more about how the BC modal will handle bouncing back and forth through sections. 

All-in-all only showing one or the other section wouldn't be too bad to just pickup on the "scrollTo" variable passed and don't render the other subsections. But since that wasn't in any A/C and since it's already focusing to their section I think that would be more of a nice to have if we have time.

packages/data/json/9.9.9.json
wardm89: Huh.. That's somewhat annoying. lol I thought that references were unique to the Building code but you're right they are not since 1.1. could refer to Division A, B or C

I can add the Volume and Division back. I'll just have to add it to the rest of the references.

packages/data/json/9.10.14.json
wardm89: Yes, to follow the same pattern would have required two separate folders for terms and building code. Rather, it should be structured in a single folder and then if we wanted to break those down with the 3 folders we can. 

What I think would be best ultimately is having the folders within `json` mimic DB data structures (tables) and have a set of types files and a single data conversion type that is the `static DB` vs having multiple exported constants to reference different things. I think it would make the refactoring to using a real DB a little less heavy of a refactor.

Since moving the files to 3 separate subfolders within json would be a simple change I'm thinking we could just do that at any time if we need more granularity for the POC.

packages/ui/src/modal-side/ModalSide.css
wardm89: Yeah one of the designs show H3/H4 for the headers (Which I have) and the other shows H4/H5 which is slightly smaller. Which is the one I had before but when I checked I saw the larger ones so I updated it lol. So which one is right? I'm thinking the smaller one is better as you suggested?

packages/ui/src/modal-glossary-content/ModalGlossaryContent.tsx
wardm89: I'm thinking each article being an article seems to align the most with the example you linked the most.

packages/data/json/9.10.14.json
nicholaslyon-dt: Yeah I originally thought of using something like [JSON Schema](https://json-schema.org/) or [JSON Type Definition](https://jsontypedef.com/) to help with some structures and types but I didn't have experience with either and didn't want to add it as another new thing to learn for this project. I never know how much to make this a POC (as it says in the repo description) vs the need for something that is 'production ready'.

packages/ui/src/modal-side/ModalSide.css
nicholaslyon-dt: From what I can tell It's a responsive thing. At desktop they use the H3/H4 styles and on mobile they use H4/H5. That's what the responsive breakdown [Templates page](https://www.figma.com/design/fABseEsz7QvzaUXz9lRyPw/BCBC---Design-Discovery?node-id=770-7499&t=bsI0kbxLlgQMEgMq-4) shows

packages/ui/src/modal-glossary-content/ModalGlossaryContent.tsx
nicholaslyon-dt: Cool. Let's do that! Because it also says articles can be nested so there could be an `<article>` at the 9.9.9 level which contains the Article `<article>`s.

packages/ui/src/modal-building-code-content/ModalBuildingCodeContent.tsx
```diff
@@ -0,0 +1,227 @@
...
+    );
+  };
+
+  const renderSentences = (sentences: SentenceType[]) => {
+    return (
+      <div className="ui-ModalBuildingCodeContent--SectionLine">
+        <div className="ui-ModalSide--SectionContent">
+          <ol type="1">
+            {sentences.map((sentence, index) => (
+              <section
```
nicholaslyon-dt, line 183: I'm not sure we can have the `<section>` elements as children of the `<ol>`'s semantically. It looks like it might take different CSS solutions to get the highlighting style to work correctly then. I would also think we don't want these to be `≤section>` elements. `<ol>` and their `<li>`'s already have an appropriate amount of semantic purpose here. This might be good to add to the polish ticket.

packages/ui/src/modal-building-code-content/ModalBuildingCodeContent.tsx
```diff
@@ -0,0 +1,227 @@
...
+        ))}
+      </ol>
+    );
+  };
+
+  const renderClauses = (clauses: SubClauseType[]) => {
+    return (
+      <ol type="a" className="ui-ModalBuildingCodeContent--List">
+        {clauses.map((data, index) => (
+          <div
```
nicholaslyon-dt, line 48: Same here as with the `<section>` elements. The `<div>`s should be inside of the `<li>`s semantically.

nicholaslyon-dt (commented):
2 possible updates to add to the polish ticket. I already added CSS notes to the polish ticket as well.

packages/ui/src/modal-building-code-content/ModalBuildingCodeContent.tsx
wardm89: Some of this will be addressed in the immediately following PR. Also, I do need some sort of html wrapper for the list items or else the ordered list will not have numbers. If I do the reference on the list item itself then the numbers go away. 

If you want to play around with it during the polishing phase I think that would be fine.

nicholaslyon-dt: Yeah I have some ideas, maybe I'll play around with it! The auto list item numbers are annoying.

PR #55, opened 2024-07-03
HOUSNAV-31: Step Tracker

nicholaslyon-dt:
[HOUSNAV-31](https://hous-bssb.atlassian.net/browse/HOUSNAV-31)

## Overview & Purpose
Add a step tracker to the display the matches the design at all screen sizes and appropriately reflects the state of the application steps.

## Summary of Changes

- Created StepTracker component
- Created `farthestItemId` in the store to reflect the current farthest item the user is on while they may navigate back to previously answered questions
- Added logic to navigation store to figure out the state of each question in the Step Tracker display
- Updated `parseStringToComponents` to account for case where we want to show questionText without any included HTML markup
- Updated section IDs in JSON to be prefixed with a number so they stay in order when turned into an Array
- Added some icons

## Side Effects
Re-named some some CSS variables to more appropriately reflect their values when they differentiate from the imported BC Design Library values

## Testing
Navigate through 9.9.9 and see the updated Step Tracker status based on questions that are complete and skipped and sections that are complete and skipped.

## Checklist

### DIFF
--- a/apps/web/components/walkthrough/Walkthrough.tsx
+++ b/apps/web/components/walkthrough/Walkthrough.tsx
@@ -49,7 +49,7 @@ const Walkthrough = observer((): JSX.Element => {
 
   return (
     <div className="web-Walkthrough" data-testid={TESTID_WALKTHROUGH}>
-      <div className="web-Walkthrough--Content">
+      <section className="web-Walkthrough--Content">
         {currentResult ? (
           <Result displayMessage={currentResult.resultDisplayMessage} />
         ) : (
@@ -63,10 +63,10 @@ const Walkthrough = observer((): JSX.Element => {
           </Form>
         )}
         <WalkthroughFooter />
-      </div>
-      <div className="web-Walkthrough--StepTracker">
+      </section>
+      <section className="web-Walkthrough--StepTracker">
         <StepTracker />
-      </div>
+      </section>
     </div>
   );
 });
--- a/apps/web/components/walkthrough/Walkthrough.tsx
+++ b/apps/web/components/walkthrough/Walkthrough.tsx
@@ -10,6 +10,7 @@ import { ID_QUESTION_FORM } from "@repo/constants/src/ids";
 import Question from "../question/Question";
 import WalkthroughFooter from "../walkthrough-footer/WalkthroughFooter";
 import Result from "../result/Result";
+import StepTracker from "../step-tracker/StepTracker";
 import { useWalkthroughState } from "../../stores/WalkthroughRootStore";
 import "./Walkthrough.css";
 
@@ -48,7 +49,6 @@ const Walkthrough = observer((): JSX.Element => {
 
   return (
     <div className="web-Walkthrough" data-testid={TESTID_WALKTHROUGH}>
-      <div className="web-Walkthrough--SideNav"></div>
       <div className="web-Walkthrough--Content">
         {currentResult ? (
           <Result displayMessage={currentResult.resultDisplayMessage} />
@@ -63,6 +63,9 @@ const Walkthrough = observer((): JSX.Element => {
           </Form>
         )}
         <WalkthroughFooter />
+      </div>
+      <div className="web-Walkthrough--StepTracker">
+        <StepTracker />
       </div>
     </div>
   );
--- a/apps/web/stores/NavigationStore.ts
+++ b/apps/web/stores/NavigationStore.ts
@@ -166,6 +166,7 @@ export class NavigationStore {
     );
   };
 
+  // an item (aka question) is skipped if it was not answered, and the section is complete, OR we are past the item in the section
   itemWasSkipped = (
     itemId: string,
     sectionId: string,
@@ -187,6 +188,7 @@ export class NavigationStore {
     );
   };
 
+  // helper function used in sectionIsComplete and sectionWasSkipped for determining if we are past the section
   pastSection = (sectionId: string) => {
     if (!this.farthestItemId) return false;
     const sections = Object.entries(this.rootStore.walkthroughData.sections);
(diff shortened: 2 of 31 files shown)


laurenrevere:
I know this is a draft, but here are some small questions/things I noticed after taking a look:

- m^2 isn’t displayed correctly.
- Any ideas what we might want to do with P23 and P24 here? Since the question takes its value from P21, would we want to surface that in the sidebar or pick a placeholder word to use here?
Here is what is currently displayed: “Is the served by a single exit stairway or ramp?” and "Is the no more than 1.5m above adjacent ground level?"

Maybe using something like the word "feature"? Might be a question for the client.


Note: the wording of P03 will be fixed in https://hous-bssb.atlassian.net/browse/HOUSNAV-69.


Side note: the functionality is looking great so far!

nicholaslyon-dt:
@laurenrevere Good catches! Updated to not skip ALL markup when skipping components (for m^2 case). Added the special case for answer value fill in as well with a default to _____. How do you think that looks/feels?

**Initial**
Screenshot 2024-07-08 at 10 20 58 AM

**Answered**
Screenshot 2024-07-08 at 10 21 10 AM

laurenrevere:
> to not skip ALL markup when skipping components (for m^2 case). Added the special case for answer value fill in as well with a default to _____. How do you think that looks/feels?

Great updates!

packages/ui/src/variables.css
```diff
@@ -230,6 +237,22 @@
...
+  --walkthrough-card-border-default: 1px solid
+    var(--surface-color-border-default);
+  --walkthrough-card-title-text-decoration-hidden: underline 1px #00000000;
+  --walkthrough-card-title-text-decoration-color-transition: underline 1px;
+  --walkthrough-card-title-text-underline-offset: 0.2em;
+  --step-tracker-heading-padding-left: var(--layout-padding-xlarge);
+  --step-tracker-padding-top: var(--layout-padding-xsmall);
+  --step-tracker-padding-left: var(--layout-padding-medium);
+  --step-tracker-padding: var(--step-tracker-padding-top)
+    var(--step-tracker-padding-left);
```
wardm89, line 253: No issue just wondering if it's your intent to do this going forward rather than just using the original variables?

apps/web/stores/NavigationStore.ts
```diff
@@ -143,4 +157,58 @@ export class NavigationStore {
...
+
+  itemIsComplete = (itemId: string) => {
+    return (
+      this.rootStore.answerStore.answers[itemId] !== undefined &&
+      (this.currentItemId !== itemId ||
+        (this.currentItemId === itemId && !this.nextButtonIsDisabled))
+    );
+  };
+
+  itemWasSkipped = (
```
wardm89, line 170: I think some comments on these for `itemWasSkipped`, `pastSection`, and `sectionIsComplete` would be helpful. Like a higher level "...checks if current item aka question is behind the farthest question answered; then it was skipped...". Something along those lines for each.

wardm89 (approved):
Approving since the only change requests are some comments. 

Update: Requesting changes after I just noticed the warning.

wardm89 (changes requested):
This is logging to the console after every question.
Screenshot 2024-07-09 at 12 59 57 PM

packages/data/json/9.10.14.json
```diff
@@ -3,10 +3,10 @@
     "title": "Spatial Separation Between Buildings",
     "subtitle": "Volume II - 9.10.14.1-4",
     "description": "Ensure a building’s separation passes code.",
-    "startingSectionId": "intro"
+    "startingSectionId": "1_intro"
```
jbook24, line 6: Why are we adding a 1_ and 2_ in front of the sectionIds?

apps/web/components/walkthrough/Walkthrough.tsx
```diff
@@ -48,7 +49,6 @@ const Walkthrough = observer((): JSX.Element => {
 
   return (
     <div className="web-Walkthrough" data-testid={TESTID_WALKTHROUGH}>
```
jbook24, line 51: Could we use Html5 semantics instead of nested <divs> ?

jbook24:
> This is logging to the console after every question. Screenshot 2024-07-09 at 12 59 57 PM

It looks like it is happening only for P21 and in this function? 

image

jbook24 (approved):

packages/ui/src/variables.css
nicholaslyon-dt: Yeah it was just because of how they had to change between mobile and desktop overall. I'm not sure I would do it this way unless I had to.

apps/web/stores/NavigationStore.ts
nicholaslyon-dt: Added some comments! LMK what you think.

nicholaslyon-dt:
> This is logging to the console after every question. Screenshot 2024-07-09 at 12 59 57 PM

Ahh wasn't really a warning we needed. Refactored `getQuestionAnswerValueDisplay` and `getAnswerValueDisplay`.

packages/data/json/9.10.14.json
nicholaslyon-dt: This is to take advantage of the ordering of the sections. When using Object.keys to turn the object of sections into an array to iterate over for display in order, I wanted them to be in a specific order. "number-like" keys use ascending order.

apps/web/components/walkthrough/Walkthrough.tsx
nicholaslyon-dt: Good call. I hadn't paid attention to the overall structure in a while. Moved the step tracker `<section>` up into the walkthrough overall and made the `web-Walkthrough--Content` it's own `<section>`. I left this overall wrapper as a `<div>` because it's just there for styling, this component is already wrapped in a `<main>` element from the root layout and I don't think there is a better way to structure that. Thoughts?

wardm89 (approved):

PR #56, opened 2024-07-03
HOUSNAV-105: Add Workflow 1 Tests

laurenrevere:
[HOUSNAV-105](https://hous-bssb.atlassian.net/browse/HOUSNAV-105)

## Overview & Purpose
Add test flows that cover all questions and all results.

## Summary of Changes
- Add tests that cover all questions and results.
- Add example flows that don't pass. (These need to be reviewed with the client).

## Side Effects
Remove some of the result text so that I don't have to worry about current formatting. This is still good enough to check that we're on the correct results page, but will not confirm that *all* of the results text is correct.


apps/web/cypress/fixtures/questionable-test-data.json
```diff
@@ -0,0 +1,171 @@
...
+        },
+        {
+          "question": "P13",
+          "type": "radio",
+          "answer": "false"
+        },
+        {
+          "question": "P17",
+          "type": "radio",
+          "answer": "true"
```
wardm89, line 79: After P17 `true` I'm getting R20 rather than going to P18. Which I believe is what we are confirming with this test that it "doesn't make it to R11"

Here are my answers after logging the current state after answering P17 and being on result R20 page.
Screenshot 2024-07-05 at 10 07 16 AM

apps/web/cypress/fixtures/questionable-test-data.json
```diff
@@ -0,0 +1,171 @@
...
+          },
+          {
+            "question": "P14",
+            "type": "radio",
+            "answer": "true"
+          },
+          {
+            "question": "P17",
+            "type": "radio",
+            "answer": "true"
```
wardm89, line 155: Same thing here, getting R20 instead of moving on to P18. Which is "correctly incorrect" haha

wardm89 (commented):
All the other test cases gave me the expected result.

apps/web/cypress/fixtures/questionable-test-data.json
laurenrevere: I added these flow examples to have a space to store the flows I think are giving an incorrect result. If we end up figuring out what the result really *should* be we can either update this data or run them as is to prove that our application has been fixed.

apps/web/cypress/fixtures/questionable-test-data.json
laurenrevere: @nicholaslyon-dt would love your or @wardm89's perspective on what we think the values  for these "questionable" flows here *should* be before we bring this to the ministry. I'm not sure if we are of the opinion that the feasibility report is wrong for these cases or we should interpret the logic differently. (Or if we think we are getting the correct answer from the tool and these flows aren't questionable at all.)

nicholaslyon-dt: Yeah I think we just need to bring up these flows with the ministry and ask. The feedback from the SME in an email was:

`ie. only check condition were both P(5) and P(8) are any value but true`

I forwarded y'all the mail with my set of questions and his answers if you want to take a look!

nicholaslyon-dt (approved):

apps/web/cypress/fixtures/questionable-test-data.json
```diff
@@ -0,0 +1,171 @@
...
+            "type": "radio",
+            "answer": "false"
+          },
+          {
+            "question": "P13",
+            "type": "radio",
+            "answer": "true"
+          },
+          {
+            "question": "P14",
```
laurenrevere, line 148: Note to self: I'm not 100% sure it makes sense to be able to make P12 false and P14 true.

apps/web/cypress/fixtures/questionable-test-data.json
```diff
@@ -0,0 +1,171 @@
...
+        },
+        {
+          "question": "P3",
+          "type": "checkbox",
+          "answer": "1"
+        },
+        {
+          "question": "P6",
+          "type": "radio",
+          "answer": "true"
```
laurenrevere, line 54: Note to self: is it possible for a dwelling unit to be above another dwelling unit if it has a basement?

apps/web/cypress/fixtures/questionable-test-data.json
laurenrevere: One of the weird things about this flow is that both "yes" and "no" to P17 have the same result.

PR #67, opened 2024-07-12
HOUSNAV-94: Workflow 2 UI - Part 2

nicholaslyon-dt:
[HOUSNAV-94](https://hous-bssb.atlassian.net/browse/HOUSNAV-94)

## Overview & Purpose
Finish out updates to the UI and logic for Workflow 2

## Summary of Changes
Refactored AnswerStore helper function `answerValuesAreNotEqual` to be in AnswerStore and not variableItem logic file
Updated AnswerStore to store variables the supplied `variableName` and not just the `currentId`
Updated questionHistory to store items with `questionId` and `answerVariableId`
Separated out some function from `set currentItemId` to make it more simple and facilitate new questionHistory updates
Added calculations
Refactored shared logic types to sharedLogic file
Added new `variableType`s number and copy

## Side Effects
Possible Workflow 1 regression with some refactoring

## Testing
Test Workflow 2!

## Checklist

### DIFF
--- a/apps/web/stores/AnswerStore.ts
+++ b/apps/web/stores/AnswerStore.ts
@@ -3,14 +3,11 @@ import { makeAutoObservable } from "mobx";
 // local
 import { WalkthroughRootStore } from "./WalkthroughRootStore";
 import {
-  QuestionVariableData,
   PropertyNameVariableToSet,
+  QuestionVariableData,
 } from "@repo/data/useWalkthroughData";
-import {
-  answerValuesAreNotEqual,
-  getVariableItemValue,
-} from "../utils/logic/variableItem";
-import { isNumber, isObject, isString } from "../utils/typeChecking";
+import { getVariableItemValue } from "../utils/logic/variableItem";
+import { isArray, isNumber, isObject, isString } from "../utils/typeChecking";
 
 export type AnswerTypes = string | string[] | Record<string, string> | number;
 export type AnswerState = Record<string, AnswerTypes>;
@@ -21,6 +18,25 @@ export type AnswerToCheckValueFn = (
 export const DEFAULT_ANSWER_VALUE_MULTI_CHOICE = null;
 export const DEFAULT_ANSWER_VALUE_NUMBER_FLOAT = NaN;
 export const DEFAULT_ANSWER_VALUE_MULTI_CHOICE_MULTI: string[] = [];
+
+export const answerValuesAreNotEqual = (
+  newAnswerValue: AnswerTypes,
+  currentAnswerValue: AnswerTypes,
+): boolean => {
+  if (isArray(newAnswerValue) && isArray(currentAnswerValue)) {
+    return !newAnswerValue.every((value) => currentAnswerValue.includes(value));
+  } else if (isString(newAnswerValue) && isString(currentAnswerValue)) {
+    return newAnswerValue !== currentAnswerValue;
+  } else if (isNumber(newAnswerValue) && isNumber(currentAnswerValue)) {
+    return newAnswerValue !== currentAnswerValue;
+  } else if (isObject(newAnswerValue) && isObject(currentAnswerValue)) {
+    return !Object.entries(newAnswerValue).every(
+      ([key, value]) => currentAnswerValue[key] === value,
+    );
+  }
+
+  throw new Error("Invalid types");
+};
 
 export class AnswerStore {
   rootStore: WalkthroughRootStore;
@@ -47,7 +63,7 @@ export class AnswerStore {
       this.getAnswerToCheckValue,
     );
 
-    const currentItemValue = this.answers[currentId];
+    const currentItemValue = this.answers[variableToSet.variableName];
 
     // This handles the case when a user has gone back and changed an answer
     // We need to erase all future answers because the flow could be different now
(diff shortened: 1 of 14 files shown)


apps/web/stores/AnswerStore.ts
```diff
@@ -22,6 +19,25 @@ export const DEFAULT_ANSWER_VALUE_MULTI_CHOICE = null;
...
+    return newAnswerValue !== currentAnswerValue;
+  } else if (isNumber(newAnswerValue) && isNumber(currentAnswerValue)) {
+    return newAnswerValue !== currentAnswerValue;
+  } else if (isObject(newAnswerValue) && isObject(currentAnswerValue)) {
+    return !Object.entries(newAnswerValue).every(
+      ([key, value]) => currentAnswerValue[key] === value,
+    );
+  }
+
+  throw new Error("Invalid types");
```
wardm89, line 38: Is this being handled up the chain with a try catch?

apps/web/utils/calculations.ts
```diff
@@ -0,0 +1,89 @@
...
+              item as VariableValueToSetCalculation,
+            );
+          },
+          answerValueToUse,
+        );
+      }
+
+      throw new Error(
+        "variableValueToSetItemCalculations must exist when variableValueCalculationType is Then",
+      );
```
wardm89, line 30: Same with these, are they handled or do they bomb the application if they occur?

apps/web/utils/logic/variableItem.ts
```diff
@@ -77,19 +113,191 @@ export const getVariableItemValueObject = (
...
-  } else if (isObject(newAnswerValue) && isObject(currentAnswerValue)) {
-    return !Object.entries(newAnswerValue).every(
-      ([key, value]) => currentAnswerValue[key] === value,
+export const getVariableItemValueNumber = (
+  variableToSet: VariableToSet,
+  getAnswerToCheckValue: AnswerToCheckValueFn,
+): number => {
+  if (isNumber(variableToSet.variableValue)) {
+    return variableToSet.variableValue;
+  } else if (isArray(variableToSet.variableValue)) {
```
wardm89:
```suggestion
const { variableValue } = variableToSet;
  if (isNumber(variableValue)) return variableValue;}

  if (!isArray(variableValue)) {
    throw new Error(
      `getVariableItemValueNumber: variableToSet.variableValue must be a number or an array, got ${typeof variableValue}`,
    );
  }
```

Small thing if you want to add this but we could do the base case type checking first to reduce nesting.
IF you added this change then I'd still add a generic thrown error after the for loop if somehow it makes it past that.

apps/web/utils/logic/variableItem.ts
```diff
@@ -77,19 +113,191 @@ export const getVariableItemValueObject = (
...
+  throw new Error(
+    `getVariableItemValueString: variableToSet.variableValue must be a number, or array, got ${typeof variableToSet.variableValue}`,
+  );
+};
+
+export const variableValueLogicItemIsTrue = (
+  variableValueLogicItem: VariableValueLogic,
+  getAnswerToCheckValue: AnswerToCheckValueFn,
+) => {
+  // if fallback item is found, set variableValue to fallback item
```
wardm89:
```suggestion
) => {
  const {
    variableValueLogicType,
    valuesToCheck,
    answerToCheck,
    answerValue,
    answerValues,
  } = variableValueLogicItem;
  // if fallback item is found, set variableValue to fallback item
```

This would benefit a lot by deconstructing here and some other places.

wardm89 (changes requested):
Looks good overall. It's hard not to do so much nesting when doing logic like this but deconstructing can help make it more readable. 

Also, let me know about the error handling stuff. Are these errors possible to even test manually to verify they don't bomb the application?

apps/web/stores/AnswerStore.ts
nicholaslyon-dt: Yeah, this is only called from `handleVariableItem` which is only called from `handleForwardNavigation` which is only called recursively in `handleVariableItem` or itself OR in the try catch in `Walkthrough.tsx`

apps/web/utils/calculations.ts
nicholaslyon-dt: They're just father down the chain. They're caught! (thanks for asking these questions, good to verify!)

apps/web/utils/logic/variableItem.ts
nicholaslyon-dt: Looking at this made me realize we don't even need to pass the whole `variableToSet` object. We just need the `variableValue` property. Refactored for that. Also I tend to like my errors at the bottom for organization and readability so I'd want to leave it there. LMK what you think!

apps/web/utils/logic/variableItem.ts
nicholaslyon-dt: Refactored! I tried to make it JIT so we're not de-structuring before we need to. LMK what you think!

wardm89 (approved):
LGTM Thanks for refactoring that!

apps/web/utils/logic/variableItem.ts
wardm89: Yeah for cases where it's a loop or something that may hit the bottom it's good to have the error case at the bottom but I like having the base cases at the top for organization and readability hahaha But yeah that's fine, not a big difference.

PR #68, opened 2024-07-12
HOUSNAV-20: Download Results To PDF

wardm89:
[HOUSNAV-20](https://hous-bssb.atlassian.net/browse/HOUSNAV-20)

## Overview & Purpose

Create a component to print the results of the walkthrough to PDF.

## Summary of Changes

- Added `PdfResultDownload` component 
- Added `PrintContent` component which will be used for all the hidden content html we want to print
- Updated `getQuestionAnswerValueDisplay` to return user friendly values for more answer types
- Added `parseComponentToPlainText` which removes the html from components so the pdf text is clean.
- Added `findBuildingCodeByNumberReference` which will return the relevant building code section a given reference number
    - This will be used later to display the building code content
- Added `findSectionTitleByQuestionId` which is used to get the display friendly title for a question  
- Added `DownloadIcon` component
## Side Effects

- I did some preliminary testing merging this with Workflow 2 but once that's merged this may need to be re-tested to make sure the pdfs print out properly for all of workflow 2. 

## Testing

- Known variations from design
    1. Question Id  ie. "P(1)" not displayed.
    2. Display intro questions without building code column since there is no reference
    3. Not displaying logic at bottom of answer
    4. Currently only showing the building code reference and not the whole code. Will be completed in [HOUSNAV-97](https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3?selectedIssue=HOUSNAV-97)
    5. Not displaying the possible answers, only the answers the user gave. 
        - This actually increases the complexity a good bit because of the logic to account for the possible answers and then get that into a format that can be printed with the selected answer highlighted. I'd make that it's own ticket if we want it during the last week or phase 3. [HOUSNAV-123 "Display Possible Answers In PDF Download"](https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3/backlog?selectedIssue=HOUSNAV-123)
    7. Not showing a building code section at the bottom, only the right if the question references the code.
    8. Width of tables may be less if not enough content is there. I'm looking to resolve this in [HOUSNAV-97](https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3?selectedIssue=HOUSNAV-97) since there seems to be a conflicting css class that I haven't found yet. Don't want to hold this pr up for that right now. 

## Notes & Resources

Screenshot 2024-07-12 at 4 35 50 PM
Screenshot 2024-07-12 at 4 36 06 PM
Screenshot 2024-07-12 at 4 36 19 PM
Screenshot 2024-07-12 at 4 36 39 PM
Screenshot 2024-07-15 at 9 19 21 AM
Screenshot 2024-07-12 at 4 37 30 PM
Screenshot 2024-07-12 at 4 37 38 PM

## Checklist


packages/ui/src/pdf-result-download/PdfResultDownload.css
```diff
@@ -0,0 +1,106 @@
...
+
+.ui-PdfResultDownload-imageContainer {
+  position: relative;
+  transition: 0.5s ease;
+  background-color: var(--pdf-result-download-container-image-container);
+  width: 229px;
+  height: 190px;
+  margin: 1px; /* prevents 1px of the 2px "focus" border from being covered */
+  border-radius: var(--pdf-result-download-image-border-radius);
+  border-color: 1px solid white;
```
nicholaslyon-dt: 2 notes here. 

1. `border-color` is an invalid prop with the full border property.
2. I think we need an `overflow: hidden` on this class so it hides the corners of image.

nicholaslyon-dt: And if you do the `hidden`, you'll need to move the right border from the image to here.

packages/ui/src/pdf-result-download/PdfResultDownload.tsx
```diff
@@ -0,0 +1,67 @@
...
+  "data-testid": testid = "",
+  ...props
+}: PdfResultDownloadProps) {
+  const handleDownload = () => {
+    window.print();
+  };
+
+  return (
+    <>
+      <ReactAriaLink
```
nicholaslyon-dt: Since we don't have an href and aren't navigating here I would tend to lean towards a button here vs a link. I could be wrong though. However, if you do that, you'd have to recreate the `button` element inside of this just in styles. Since React Aria doesn't render an `<a>` tag here since it's not given an href this might be fine but we might want to see if we can explicitly set the aria-role to button instead of link.

packages/ui/src/pdf-result-download/PdfResultDownload.css
```diff
@@ -0,0 +1,106 @@
...
+  border-right: 1px solid var(--surface-color-border-default);
+}
+
+.ui-PdfResultDownload-imageContainer {
+  position: relative;
+  transition: 0.5s ease;
+  background-color: var(--pdf-result-download-container-image-container);
+  width: 229px;
+  height: 190px;
+  margin: 1px; /* prevents 1px of the 2px "focus" border from being covered */
```
nicholaslyon-dt: Ideally we remove this and move the outline. We did this for the button, you can see it in the `button.css`. React Aria nicely adds a data prop to style when the focus should be visible. `[data-focus-visible]`. Thoughts on copying the button focus style for this large button?

packages/ui/src/print-content/PrintContent.tsx
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    return grouped;
+  };
+
+  // Get the question data from the question history for the PDF
+  const questionPdf = questionHistory
+    .map((questionId): QuestionData | undefined => {
+      const question = walkthroughData.questions[questionId];
+      if (!question || !("questionText" in question)) {
```
nicholaslyon-dt: Can we use the `getQuestionAsDisplayType` helper function here? The display type question currently always has "questionText".

I'm also just realizing that means `getQuestionTextByQuestionId` could be refactored to call `getQuestionAsDisplayType` as well...

apps/web/utils/string.tsx
```diff
@@ -177,6 +177,26 @@ export const parseStringToComponents = (
   return parse(html, options);
 };
 
+export const parseComponentToPlainText = (component: ReactNode): string => {
```
nicholaslyon-dt: This may already be what the `getStringFromComponents` function below this does. Did that not work for this case?

apps/web/stores/WalkthroughRootStore.ts
```diff
@@ -181,20 +181,63 @@ export class WalkthroughRootStore {
...
       );
+
       const displayValue =
-        answerValue?.answerValueDisplay ?? answerValue?.answerDisplayText;
+        matchedAnswer?.answerValueDisplay ?? matchedAnswer?.answerDisplayText;
       return displayValue ?? "";
     } else if (isNumber(answer)) {
       return answer.toString();
+    } else {
+      const cleanAnswer = toJS(answer);
```
nicholaslyon-dt: I haven't been doing this and it's worked fine for me. Did it not work for you?

apps/web/stores/WalkthroughRootStore.ts
```diff
@@ -181,20 +181,63 @@ export class WalkthroughRootStore {
...
+
+      if (isArray(cleanAnswer) && "possibleAnswers" in question) {
```
nicholaslyon-dt: Let's use the var form of this string `PropertyNamePossibleAnswers`

packages/ui/src/icon/icons/DownloadIcon.tsx
```diff
@@ -0,0 +1,19 @@
+export default function DownloadIcon() {
+  return (
+    <svg
+      width="21"
+      height="20"
+      viewBox="0 0 21 20"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
```
nicholaslyon-dt, line 10: Can we update this to match the other svg icon classes with a passed title and id? And then also update the icon.test.tsx to include this new icon type? Github copilot just writes them for you at this point.

packages/data/json/9.9.9.json
```diff
@@ -3,7 +3,8 @@
     "title": "Egress from Dwelling Units",
     "subtitle": "Volume II - 9.9.9.",
     "description": "Ensure your dwelling unit has exits or egress doors that pass code.",
-    "startingSectionId": "1_intro"
+    "startingSectionId": "1_intro",
+    "walkthroughTitle": "9.9.9. Egress from Dwelling Units"
```
nicholaslyon-dt, line 7: Do we want to just make this the number? Since we're just copying the `info.title` here in both cases. Then construct it?

packages/data/json/9.10.14.json
```diff
@@ -3,7 +3,8 @@
     "title": "Spatial Separation Between Buildings",
     "subtitle": "Volume II - 9.10.14.1-4",
     "description": "Ensure a building’s separation passes code.",
-    "startingSectionId": "1_intro"
+    "startingSectionId": "1_intro",
+    "walkthroughTitle": "9.10.14. Spatial Separation Between Buildings"
```
nicholaslyon-dt: Should we add the `1-4` here as well since this doesn't cover 9.10.14.5?

nicholaslyon-dt (changes requested):
Overall this is looking dope! I love it. Few comments about styling and semantics for the button card, then just some code questions.

packages/ui/src/pdf-result-download/PdfResultDownload.css
wardm89: Ahh thanks, I was trying something out and forgot to take this out.

apps/web/stores/WalkthroughRootStore.ts
wardm89: You're right that was necessary for logging but not for the code to function.

apps/web/utils/string.tsx
wardm89: Wow how did I miss that, yes it does the same thing just a little differently by checking the types with a switch instead.

packages/data/json/9.10.14.json
wardm89: Makes sense, yes

packages/data/json/9.9.9.json
wardm89: I'm copying the pattern for the `sectionTitle` Which includes it. I think since we are manually doing this in json and it only has a use case to display them together then we should just keep them together here.

packages/ui/src/pdf-result-download/PdfResultDownload.css
wardm89: Done, that makes it look better overall I think

wardm89:
Had some failures for a type matching issue I didn't catch locally since it was still working.

packages/ui/src/icon/Icon.test.tsx
nicholaslyon-dt, line 1: Thanks for this 😂.

nicholaslyon-dt (commented):
2 small things, but other than late this looks great!

1. I'm not seeing the bottom border below the image.
Screenshot 2024-07-16 at 12 35 02 PM

2. On mobile it goes off the screen. This is at 375px wide.
Screenshot 2024-07-16 at 12 36 29 PM

packages/ui/src/icon/Icon.test.tsx
wardm89: Yeah when I was adding it, I was like "Are all these the same?" haha simple update and will make it easier to add an icon later

nicholaslyon-dt (approved):

PR #69, opened 2024-07-16
HOUSNAV-97: Add Building Code to PDF download

wardm89:
[HOUSNAV-97](https://hous-bssb.atlassian.net/browse/HOUSNAV-97)

## Overview & Purpose

Add full building code for relevant references to PDF printout.

## Summary of Changes

- Added `printData` to buildingCodeContent to enable smaller portions of building code to be printed. 
- Updated groupings for `groupedQuestions` for the pdf in order to group by `section` then group within the section by `reference`. 
- Added a new table at the bottom of the PDF printout to display the building code data as it's own table. 
- CSS updates ensuring spacing and width for the table make sense. 

## Testing

- Building code printout sections are broken down into the largest encompassing section, article, sentence for the group of references for the section.
    - In the case of 10.14 there is a reference that references the whole building code section so that will mean that for 10.14 the whole subsection will be printed. 

- Known variations from design
    1. Question Id  ie. "P(1)" not displayed.
    2. Display intro questions without building code column since there is no reference
    3. Not displaying logic at bottom of answer
    4. Currently only showing the building code reference only to the right of the relevant questions in groups; then displaying full building code portions at the bottom. 
    5. Not displaying the possible answers, only the answers the user gave. 
        - This actually increases the complexity a good bit because of the logic to account for the possible answers and then get that into a format that can be printed with the selected answer highlighted. I'd make that it's own ticket if we want it during the last week or phase 3. [HOUSNAV-123 "Display Possible Answers In PDF Download"](https://hous-bssb.atlassian.net/jira/software/c/projects/HOUSNAV/boards/3/backlog?selectedIssue=HOUSNAV-123)

## Notes & Resources

Screenshot 2024-07-16 at 5 05 29 PM
Screenshot 2024-07-16 at 4 13 31 PM

## Checklist


jbook24 (approved):

jbook24:
Is the title centered? 
image

wardm89:
> Is the title centered? image

No the title is on the left similar to the design and modal

packages/ui/src/print-content/PrintContent.css
```diff
@@ -2,29 +2,26 @@
...
 }
 
-.ui-printContent--questionTable td,
-.ui-printContent--questionTable th {
-  break-inside: avoid;
-}
-
-.ui-printContent--questionTable {
+.ui-printContent--table {
   display: block;
```
nicholaslyon-dt: Had had changed this in my PR because I noticed it while doing the results UI. If the question text and responses for all questions in a section aren't long enough, the table didn't stretch full width. You can see it in this example here. If you just remove this and leave the table display as `display: table` it should work. Thoughts?

Screenshot 2024-07-18 at 11 03 12 AM

nicholaslyon-dt: Altho it doesn't seem to be doing it now so maybe it's not needed? Just curious why this was set to block to begin with a guess. I wouldn't think it was needed.

nicholaslyon-dt (approved):
Only one small question. Otherwise this looks good to me!

packages/ui/src/print-content/PrintContent.css
wardm89: Thanks for the callout, the display block isn't actually necessary. But "It's not doing it anymore" is due to the browser caching css. Which has been really fun to deal with. :D Sometime's I make a change and it doesn't work but it's because I need to refresh the cache so the browser picks up the latest css for the pdf print. 

But THIS PR should not have the inconsistent width issue like the last PR had.

nicholaslyon-dt: Ahh interesting! That would make it... very interesting. And also maybe why I was seeing it still in emulation, but not on the print dialog. Weird!

PR #81, opened 2024-07-23
HOUSNAV-128: Add Accessibility scans to all walkthrough questions and results

wardm89:
[HOUSNAV-#](https://hous-bssb.atlassian.net/browse/HOUSNAV-#)

## Overview & Purpose

Add `axe-core` accessibility testing to cypress test cases for each step and result.

## Summary of Changes

- Update `runWalkthrough` helper to run accessibility for each question and result
- Update application to correct any accessibility issues that came up
- Fixed the following issues on walkthrough questions and result pages:
    - Ensures the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds (color-contrast)
    - Heading levels should only increase by one
    - Interactive controls must not be nested

## Side Effects

Time to run tests when from about `1:20s` to about `1:50s`. Which I think is an acceptable increase to run it for each question for every test case at this point.

Before:
Screenshot 2024-07-23 at 10 12 44 AM
After:
Screenshot 2024-07-23 at 10 11 11 AM

## Testing

- e2e tests should pass. To verify they are actually running on each question you can locally change something like adding a h3 & h5 back to back which is out of order. This should trigger the accessibility test to fail.

**EXAMPLE**
If a test fails it will look like this and highlight the question in red that failed.
Screenshot 2024-07-23 at 10 26 43 AM

## Notes & Resources

https://dequeuniversity.com/rules/axe/4.9/color-contrast?application=axeAPI
https://dequeuniversity.com/rules/axe/4.9

Screenshot 2024-07-23 at 10 16 39 AM
Screenshot 2024-07-23 at 10 16 24 AM
Screenshot 2024-07-23 at 10 16 54 AM

## Checklist

### DIFF
--- a/packages/ui/src/header/Header.tsx
+++ b/packages/ui/src/header/Header.tsx
@@ -66,7 +66,7 @@ const getNavList = (onLinkClick: (href: string) => void) => {
 export default function Header({
   skipLinks,
   title = "",
-  titleElement = "span",
+  titleElement = "h1",
   logoSrc,
 }: PropsWithChildren<HeaderProps>) {
   // setup state
--- a/apps/web/components/result/Result.tsx
+++ b/apps/web/components/result/Result.tsx
@@ -27,7 +27,7 @@ export default function Result({
         <div className="u-container-walkthrough">
           <header className="web-Result--Header p-hide">
             <Icon type={"checkCircle"} />
-            <Heading level={1} className="h4">
+            <Heading level={2} className="h4">
               Walkthrough Complete
             </Heading>
           </header>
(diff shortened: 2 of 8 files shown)


packages/ui/src/header/Header.tsx
```diff
@@ -66,7 +66,7 @@ const getNavList = (onLinkClick: (href: string) => void) => {
 export default function Header({
   skipLinks,
   title = "",
-  titleElement = "span",
+  titleElement = "h1",
```
nicholaslyon-dt, line 69: I mean to make this change forever ago and completely forgot about it 😂 .

nicholaslyon-dt (approved):
I'm assuming all these tweaks were issues raised during the scanning. Looks great!

wardm89:
> I'm assuming all these tweaks were issues raised during the scanning. Looks great!

Yep all the changes were needed for the accessibility tests to pass.

PR #91, opened 2024-07-25
HOUSNAV-122: Modal accessibility updates

wardm89:
[HOUSNAV-122](https://hous-bssb.atlassian.net/browse/HOUSNAV-122)

## Overview & Purpose

Enhance the accessibility of the application by improving screen reader readouts when the modal opens and terms are highlighted.

## Summary of Changes

- Add tab indexes where necessary to improve keyboard accessiblity
- Focus on building code and terms when modal is open so screen reader can read that first
- Added hidden text that will be read after a building code or glossary section is read to inform user they can scroll or tab
- Improved modal close label

## Testing

Use a screen reader when navigating the application and check A/C in ticket
- I didn't add any new accessibility tests for the modal since they are failing and I wanted to make sure this is done before I leave.

## Checklist
- [ ] I have written or updated vitests for this work

### DIFF
--- a/apps/web/components/defined-term/DefinedTerm.tsx
+++ b/apps/web/components/defined-term/DefinedTerm.tsx
@@ -34,6 +34,7 @@ export default function DefinedTerm({
 
   const tooltipButton = (
     <Tooltip
+      tooltipLabel={`${term} select to open glossary modal`}
       tooltipContent={TooltipGlossaryData.get(tooltipTerm.toLocaleLowerCase())}
       triggerContent={button}
     ></Tooltip>
--- a/apps/web/components/defined-term/DefinedTerm.tsx
+++ b/apps/web/components/defined-term/DefinedTerm.tsx
@@ -34,7 +34,7 @@ export default function DefinedTerm({
 
   const tooltipButton = (
     <Tooltip
-      tooltipLabel={`${term} select to open glossary modal`}
+      tooltipLabel={`Select to open defined term modal`}
       tooltipContent={TooltipGlossaryData.get(tooltipTerm.toLocaleLowerCase())}
       triggerContent={button}
     ></Tooltip>
--- a/packages/ui/src/modal-building-code-content/ModalBuildingCodeContent.tsx
+++ b/packages/ui/src/modal-building-code-content/ModalBuildingCodeContent.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import React from "react";
+import React, { useRef } from "react";
 import {
   ArticleType,
   SubClauseType,
@@ -38,6 +38,7 @@ const BuildingCodeContent: React.FC<BuildingCodeContentProps> = ({
   sectionRefs = { current: {} },
   setFocusSection,
 }) => {
+  const liveRegionRef = useRef<HTMLDivElement>(null);
   const [printReference, setPrintReference] = useState<string | null>(null);
 
   useEffect(() => {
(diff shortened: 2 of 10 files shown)


apps/web/components/defined-term/DefinedTerm.tsx
```diff
@@ -34,6 +34,7 @@ export default function DefinedTerm({
 
   const tooltipButton = (
     <Tooltip
+      tooltipLabel={`${term} select to open glossary modal`}
```
nicholaslyon-dt: It's currently reading the term twice, once for the button text and once for this insertion in the label. I'd also probably call it the 'defined terms' modal as that is the title in the  modal and what the building code calls it.

packages/ui/src/modal-side/ModalSide.tsx
```diff
@@ -78,9 +78,10 @@ export default function ModalSide({
           <Dialog className="ui-ModalSide--AriaDialog">
             {({ close }) => (
               <>
-                <header className="ui-ModalSide--Header">
-                  <ButtonModalClose label="Close modal" onPress={close} />
-                </header>
+                <ButtonModalClose
```
nicholaslyon-dt, line 81: Removing the header element is probably good for accessibility, but then we need to update the styles as this moves the button location.

packages/ui/src/modal-building-code-content/ModalBuildingCodeContent.tsx
```diff
@@ -80,6 +90,7 @@ const BuildingCodeContent: React.FC<BuildingCodeContentProps> = ({
                 ? "ui-ModalSide--SectionHighlighted"
                 : ""
             }`}
+            tabIndex={0}
```
nicholaslyon-dt, line 93: We might have to figure out a different solution here as these are now just tab-able at all times, but nothing is said. It would be confusing for the user I think if they press tab a few times while going through this content with nothing being read or happening verbally.

nicholaslyon-dt (changes requested):
Few comments on things. As things stand these might be notes for me 😂 .

Overall though does the stuff read when the modal opens for you? I get the appropriate spot focused in the modal, but the content doesn't read. It just says you're in an article on a section heading.

packages/ui/src/modal-building-code-content/ModalBuildingCodeContent.tsx
wardm89: So, for me the screen reader does read if I put the tabIndex on these. It does NOT read the definition if I tab to a new definition when I don't have `tabIndex{0}`
This video shows the current state for me:
https://github.com/user-attachments/assets/3fc1c2f9-44bb-4974-ba49-63babcfbdba6

This video shows how it behaves if I remove the tabIndex:
https://github.com/user-attachments/assets/2aecb6ca-268b-47aa-a1e6-e9f0fe1126d1

packages/ui/src/modal-side/ModalSide.css
```diff
@@ -38,13 +38,7 @@
...
 }
 
-.ui-ModalSide--Header {
-  z-index: 1;
-  position: fixed;
-  top: 0;
-}
-
-.ui-ModalSide--Header .ui-ButtonModalClose.--icon {
+.ui-ModalSide--Modal .ui-ButtonModalClose.--icon {
```
nicholaslyon-dt, line 41: We still need the styles from the header here as not adding a z-index makes the button go behind the images in 9.10.14.

Screenshot 2024-07-26 at 10 10 07 AM

wardm89:
Closing since this work is combined with https://github.com/bcgov/House-Innovations-HOUSNAV/pull/93

PR #92, opened 2024-07-25
HOUSNAV-67: Accessible Disabled Button

nicholaslyon-dt:
[HOUSNAV-67](https://hous-bssb.atlassian.net/browse/HOUSNAV-67)

## Overview & Purpose
Make disabled buttons accessible to users navigating via tabbing. 

## Summary of Changes
Returned a standard HTML <button> element for disabled buttons instead of using React Aria Components
Updated styles based on changes
Updated colors to be closer to design but still accessible

## Testing
Us a walkthrough and see the ability to tab to disabled buttons in the walkthrough footer

## Checklist

### DIFF
--- a/packages/ui/src/variables.css
+++ b/packages/ui/src/variables.css
@@ -30,7 +30,7 @@
     0 25.600000381469727px 57.599998474121094px 0 #00000038; /* Primarily used for dialogs */
   --surface-color-primary-button-default: #013366;
   --surface-color-primary-button-hover: #1e5189;
-  --surface-color-primary-button-disabled: #9f9d9c;
+  --surface-color-primary-button-disabled: #edebe9;
   --surface-color-primary-danger-button-default: #ce3e39;
   --surface-color-primary-danger-button-hover: #a2312d;
   --surface-color-primary-danger-button-disabled: #edebe9;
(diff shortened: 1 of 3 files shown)


packages/ui/src/variables.css
```diff
@@ -30,7 +30,7 @@
     0 25.600000381469727px 57.599998474121094px 0 #00000038; /* Primarily used for dialogs */
   --surface-color-primary-button-default: #013366;
   --surface-color-primary-button-hover: #1e5189;
-  --surface-color-primary-button-disabled: #9f9d9c;
+  --surface-color-primary-button-disabled: #edebe9;
```
wardm89, line 33: Glad you did this. I used the axe site's tool to get as light a grey of the same hue as I could to pass the test but I guess their site is more conservative than the tests we run? idk

wardm89 (approved):
Works better for me!

packages/ui/src/variables.css
nicholaslyon-dt: Yeah I just used the WAVE chrome extension. It has a nice built in one on the page. But also the original one didn't fail via the tool, but did fail when inputting the values directly. 🤷🏻 

https://chromewebstore.google.com/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh

jbook24 (approved):

PR #119, opened 2024-08-13
HOUSNAV-164: updating result logic to be greaterThanOrEqual vs greaterThan

nicholaslyon-dt:
[HOUSNAV-164](https://hous-bssb.atlassian.net/browse/HOUSNAV-164)

## Overview & Purpose
Updating logic to match table in BCBC to be greaterThanOrEqual to

## Summary of Changes
Updated the logic for R20 in 9.10.14 to match the table in the BCBC as greater than or equal to

## Side Effects
Should be none as this only affects code directly related to R20 in 9.10.14

## Testing
Go through the walkthrough and verify cases where your limiting distance comes to 1.2, 1.5, and 2.0 gives you the correct response

## Checklist

### DIFF
--- a/packages/data/src/useWalkthroughData.ts
+++ b/packages/data/src/useWalkthroughData.ts
@@ -200,7 +200,7 @@ export enum ResultCalculationType {
 
 export enum ResultLogicTypes {
   Fallback = "fallback",
-  GreaterThan = "greaterThan",
+  GreaterThanOrEqual = "greaterThanOrEqual",
 }
 
 export interface ResultLogicItem {
--- a/apps/web/cypress/fixtures/workflow2-test-data.json
+++ b/apps/web/cypress/fixtures/workflow2-test-data.json
@@ -924,6 +924,391 @@
         }
       ],
       "result": "R30"
+    },
+    {
+      "title": "P6 = 1.5",
+      "steps": [
+        {
+          "question": "P01",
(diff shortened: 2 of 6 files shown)


packages/data/src/useWalkthroughData.ts
```diff
@@ -200,7 +200,7 @@ export enum ResultCalculationType {
 
 export enum ResultLogicTypes {
   Fallback = "fallback",
-  GreaterThan = "greaterThan",
+  GreaterThanOrEqual = "greaterThanOrEqual",
```
laurenrevere, line 203: Double checking: We don't need a normal greater than for any of the results logic types?

laurenrevere:
I added some quick UI tests to test these values E2E: In theory, it would be more efficient to just check the values instead of the UI, but this is a nice way to add these tests quickly. It also would be nicer to make a results object that contains both the result value and the result number, but wanted to make as few changes as possible getting this in. Happy to change if it feels egregious.

packages/data/src/useWalkthroughData.ts
nicholaslyon-dt: Correct! This is the only section that has "resultLogicType" in it. It's a one off for the table value retrieval in the result.

laurenrevere (approved):

PR #131, opened 2024-10-16
HOUSNAV-200: Updated Application State

nicholaslyon-dt:
[HOUSNAV-200](https://hous-bssb.atlassian.net/browse/HOUSNAV-200)

## Overview & Purpose
Update the application state to account for a walkthrough that has multiple sets of walkthrough questions and results.

## Summary of Changes
- Updated walkthrough URLS to use a query parameter
- Query parameter(s) are used to build the full walkthrough data
- Updated application store code to account for the possibility of multiple walkthroughs with different sets of questions and results

## Side Effects
All of them?

## Testing
- Make sure the current single dwelling walkthroughs work
- Make sure different configurations of multi dwelling walkthroughs allow the user to get to a result

## Notes & Resources
- The step tracker and results updates are captured in separate tickets

## Checklist

### DIFF
--- a/apps/web/components/question/Question.test.tsx
+++ b/apps/web/components/question/Question.test.tsx
@@ -1,11 +1,12 @@
 // 3rd party
 import { describe, expect, it } from "vitest";
 // repo
-import useWalkthroughTestData, {
+import {
+  useWalkthroughTestData999,
   getMultiChoiceMultipleQuestion,
   getMultiChoiceQuestion,
   getNumberFloatQuestion,
-} from "@repo/data/useWalkthroughTestData";
+} from "@repo/data/useWalkthroughsTestData";
 import {
   GET_TESTID_CHECKBOX_GROUP,
   GET_TESTID_NUMBER_FIELD,
@@ -14,7 +15,8 @@ import {
   TESTID_QUESTION_CODE_REFERENCE,
   TESTID_QUESTION_TITLE,
 } from "@repo/constants/src/testids";
-import { PropertyNameQuestionText } from "@repo/data/useWalkthroughData";
+import { PropertyNameQuestionText } from "@repo/data/useWalkthroughsData";
+import { EnumWalkthroughIds } from "@repo/constants/src/constants";
 // local
 import Question from "./Question";
 import {
@@ -27,12 +29,17 @@ describe("Question", () => {
   // renders null if no question type found
   it("Question: renders null if it doesn't get question data", () => {
     // get data
-    const walkthroughData = useWalkthroughTestData();
-    walkthroughData.info.startingSectionId = "TESTQUESTION";
+    const walkthroughData = useWalkthroughTestData999();
+    walkthroughData.walkthroughsById[
+      EnumWalkthroughIds._9_9_9
+    ].info.startingSectionId = "TESTQUESTION";
     const testQuestion = getMultiChoiceQuestion();
     testQuestion.questionData.walkthroughItemType = "TESTQUESTION";
-    walkthroughData.questions[walkthroughData.info.startingSectionId] =
-      testQuestion.questionData;
+    walkthroughData.walkthroughsById[EnumWalkthroughIds._9_9_9].questions[
+      walkthroughData.walkthroughsById[
+        EnumWalkthroughIds._9_9_9
+      ].info.startingSectionId
+    ] = testQuestion.questionData;
 
     // render component
     const { queryByTestId } = renderWithWalkthroughProvider({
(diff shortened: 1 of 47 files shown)


jbook24:
Design 
image


Current 
image


- Are we thinking the step tracker will update to reflect the current design from Figma? 
- Currently if you select "all" it only shows the new section after completing the prior

apps/web/components/question/Question.test.tsx
```diff
@@ -94,7 +103,7 @@ describe("Question", () => {
   // check question with code reference
   it("QuestionMultiChoice: renders with code reference", () => {
     // get data
-    const walkthroughData = useWalkthroughTestData();
+    const walkthroughData = useWalkthroughTestData999();
```
jbook24, line 106: As we prepare for significant uptick in use cases do we feel this approach is scalable?

apps/web/stores/AnswerStore.ts
```diff
@@ -87,19 +113,26 @@ export class AnswerStore {
   setAnswerValueOnChange = (value: AnswerTypes, questionId: string) => {
     this.eraseAnswersAfterCurrentItem();
 
-    this.rootStore.navigationStore.farthestItemId = questionId;
+    this.rootStore.navigationStore.setFarthestIds(questionId);
 
-    this.answers[questionId] = value;
+    this.setAnswerValue(
+      this.rootStore.navigationStore.currentWalkthroughId,
```
jbook24, line 119: Is it possible for this "id" to ever be out of sync?

apps/web/stores/NavigationStore.ts
```diff
@@ -50,18 +64,35 @@ export class NavigationStore {
     return "";
   }
 
+  setFarthestIds = (itemId: string, walkthroughId?: string) => {
```
jbook24, line 67: What does this mean?

apps/web/stores/NavigationStore.ts
```diff
@@ -90,44 +121,54 @@ export class NavigationStore {
 
   /*
    * Back button is disabled if:
-   * - current question is the first question
+   * - current question is the first question in the history
    */
   get backButtonIsDisabled() {
```
jbook24, line 126: Should this "back button" be enabled to allow for users to update the code selections they want to review? 

image

packages/data/json/building-types/multi-dwelling/wt-multi-dwelling-9.10.14.json
jbook24, line 1: This will be updated when the SME shares the updated version?

nicholaslyon-dt:
> Design image
> 
> Current image
> 
> * Are we thinking the step tracker will update to reflect the current design from Figma?
> * Currently if you select "all" it only shows the new section after completing the prior

@jbook24 Yes, I was planning for all updates related to the step tracker and results screen will be handled in the [step tracker](https://hous-bssb.atlassian.net/browse/HOUSNAV-199) and [results](https://hous-bssb.atlassian.net/browse/HOUSNAV-191) tickets.

apps/web/components/question/Question.test.tsx
nicholaslyon-dt: Yeah, this update was just to be more explicit here. This is just so the test reads with exactly what type of test data it is using. I don't see the need for the tests to start using all the data from different types of walkthroughs.

apps/web/stores/AnswerStore.ts
nicholaslyon-dt: I don't think so. This only happens within the context of 1 question, so the current walkthrough id wouldn't change while a user is interacting with a question. And all state is updated before any rendering happens between questions.

apps/web/stores/NavigationStore.ts
nicholaslyon-dt: It's just a helper function to update both of the stored values for the farthest question and farthest walkthrough the user has currently been to.

apps/web/stores/NavigationStore.ts
nicholaslyon-dt: No, this is what the breadcrumbs would be for. That back only ever operates within a walkthrough, it doesn't navigate to a different page.

packages/data/json/building-types/multi-dwelling/wt-multi-dwelling-9.10.14.json
nicholaslyon-dt: yep! I just have enough test data in there to make sure things work.

jbook24 (approved):

PR #141, opened 2024-11-07
HOUSNAV-193: Breadcrumbs

nicholaslyon-dt:
[HOUSNAV-193](https://hous-bssb.atlassian.net/browse/HOUSNAV-193)

## Overview & Purpose
Adding breadcrumbs to match the design

## Summary of Changes
Created Breadcrumbs component and added it to the root layout

## Testing
- Navigate to the home page - notice no breadcrumbs
- Navigate to all lower pages (building type analysis, single dwelling, multi dwelling, walkthroughs) and notice the breadcrumbs

## Checklist

### DIFF
--- a/apps/web/cypress/e2e/multi-dwelling/multi-dwelling-9.10.14-walkthrough.cy.ts
+++ b/apps/web/cypress/e2e/multi-dwelling/multi-dwelling-9.10.14-walkthrough.cy.ts
@@ -1,6 +1,5 @@
 import { URLS_GET_WALKTHROUGH } from "@repo/constants/src/urls";
 import {
-  EnumBreadcrumbIds,
   EnumBuildingTypes,
   EnumWalkthroughIds,
 } from "@repo/constants/src/constants";
@@ -8,6 +7,8 @@ import { runWalkthrough } from "../../support/helpers";
 import {
   GET_TESTID_BREADCRUMBS_BREADCRUMB,
   GET_TESTID_STEP_TRACKER_WALKTHROUGH_HEADER,
+  TESTID_BREADCRUMB_HOME,
+  TESTID_BREADCRUMB_LAST,
   TESTID_BREADCRUMBS,
 } from "@repo/constants/src/testids";
 import { walkthroughs } from "../../fixtures/multi-dwelling/multi-dwelling-9.10.14-test-data.json";
@@ -31,15 +32,11 @@ describe("multi dwelling: 9.10.14 walkthrough", () => {
 
   it("should display all parts of breadcrumbs", () => {
     cy.getByTestID(TESTID_BREADCRUMBS).should("be.visible");
+    cy.getByTestID(TESTID_BREADCRUMB_HOME).should("be.visible");
     cy.getByTestID(
-      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.HOME),
+      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBuildingTypes.MULTI_DWELLING),
     ).should("be.visible");
-    cy.getByTestID(
-      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.MULTI_DWELLING_UNIT),
-    ).should("be.visible");
-    cy.getByTestID(
-      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.LAST),
-    ).should("be.visible");
+    cy.getByTestID(TESTID_BREADCRUMB_LAST).should("be.visible");
   });
 
   it("step tracker doesn't show section headers for single section", () => {
--- a/apps/web/cypress/e2e/multi-dwelling/multi-dwelling-9.10.14-walkthrough.cy.ts
+++ b/apps/web/cypress/e2e/multi-dwelling/multi-dwelling-9.10.14-walkthrough.cy.ts
@@ -1,10 +1,15 @@
 import { URLS_GET_WALKTHROUGH } from "@repo/constants/src/urls";
 import {
+  EnumBreadcrumbIds,
   EnumBuildingTypes,
   EnumWalkthroughIds,
 } from "@repo/constants/src/constants";
 import { runWalkthrough } from "../../support/helpers";
-import { GET_TESTID_STEP_TRACKER_WALKTHROUGH_HEADER } from "@repo/constants/src/testids";
+import {
+  GET_TESTID_BREADCRUMBS_BREADCRUMB,
+  GET_TESTID_STEP_TRACKER_WALKTHROUGH_HEADER,
+  TESTID_BREADCRUMBS,
+} from "@repo/constants/src/testids";
 import { walkthroughs } from "../../fixtures/multi-dwelling/multi-dwelling-9.10.14-test-data.json";
 import { results } from "../../fixtures/results-data.json";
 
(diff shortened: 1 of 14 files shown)


packages/constants/src/constants.ts
```diff
@@ -18,6 +18,7 @@ export enum EnumBuildingTypes {
 }
 
 export const STR_BUILDING_TYPE_ANALYSIS_ID = "building-type-analysis";
+export const STR_WALKTHROUGH_ID = "walkthrough";
```
brianhoover-willowtree, line 21: Looks to follow the naming convention of the line above, but why is the value 'walkthrough' an ID? Same applies to the line above.

nicholaslyon-dt: I honestly asked myself that same question while writing it 😂. I left it the same as the line above because they have similar uses in that they are both used for the url path. The building type string was named ID originally because it was the ID of the building type walkthrough. The other walkthroughs have their own IDs (9.9.9, 9.10.14). I think it'd be fine removing the '_ID' on both if we felt it would be less confusing.

brianhoover-willowtree: I don't have a strong opinion either way, and I'm not trying to pile more work on a functioning PR. I'll consider this question closed and approve the PR.

brianhoover-willowtree (approved):

packages/constants/src/testids.ts
```diff
@@ -59,6 +59,10 @@ export const TESTID_HEADER_MOBILE_NAV = "header-mobile-nav";
 export const TESTID_HEADER_MOBILE_NAV_BUTTON = "header-mobile-nav-button";
 export const GET_TESTID_HEADER_NAV_ITEM = (title: string) =>
   `header-nav-item-${title}`;
+export const TESTID_BREADCRUMBS = "breadcrumbs";
+export const GET_TESTID_BREADCRUMBS_BREADCRUMB = (id: string) =>
+  `breadcrumb-${id}`;
+export const TESTID_BREADCRUMB_LAST = "breadcrumb-Last";
```
stechkev: capitalized "L" in last to match the rest of the IDs

apps/web/cypress/e2e/single-dwelling/single-dwelling-landing.cy.ts
stechkev, line 1: Changing from .exists to .be.visible to better test UI elements

apps/web/cypress/e2e/multi-dwelling/multi-dwelling-landing.cy.ts
stechkev, line 1: Changing from .exists to .be.visible to better test UI elements

apps/web/cypress/e2e/single-dwelling/single-dwelling-9.9.9-results.cy.ts
```diff
@@ -23,6 +24,8 @@ describe("single dwelling: 9.9.9 walkthrough results", () => {
   });
 
   it("should display correct ui elements", () => {
+    cy.getByTestID(TESTID_BREADCRUMBS).should("be.visible");
```
stechkev, line 27: Adding check for the breadcrumbs as a whole exist. we test each individual section once in the multi-dwelling-walkthrough

apps/web/cypress/e2e/multi-dwelling/multi-dwelling-combined-results.cy.ts
```diff
@@ -26,6 +27,8 @@ describe("multi dwelling: 9.9.9 and 9.10.14 walkthrough results", () => {
   });
 
   it("should display correct ui elements", () => {
+    cy.getByTestID(TESTID_BREADCRUMBS).should("be.visible");
```
stechkev, line 30: Adding check for the breadcrumbs as a whole exist. we test each individual section once in the multi-dwelling-walkthrough

apps/web/cypress/e2e/multi-dwelling/multi-dwelling-9.10.14-walkthrough.cy.ts
```diff
@@ -24,6 +29,19 @@ describe("multi dwelling: 9.10.14 walkthrough", () => {
...
+      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.HOME),
+    ).should("be.visible");
+    cy.getByTestID(
+      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.MULTI_DWELLING_UNIT),
+    ).should("be.visible");
+    cy.getByTestID(
+      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.LAST),
+    ).should("be.visible");
+  });
+
```
stechkev: checking each section of the breadcrumbs at a spot where we have a lot of sections. I only did this once because if it were to fail I only want it to fail once, in a test that's specific to the breadcrumb sections

packages/constants/src/constants.ts
```diff
@@ -17,7 +17,15 @@ export enum EnumBuildingTypes {
   MULTI_DWELLING = "multi-dwelling",
 }
 
+export enum EnumBreadcrumbIds {
```
nicholaslyon-dt: @stechkev I'm not necessarily against this, but I'm not sure we need it. This just duplicates the other strings we have for page titles. In the unit tests I'm using the helper function of getting the page name by using the URL segment. And the url segments are just the building type enums.

```javascript
GET_TESTID_BREADCRUMBS_BREADCRUMB(GET_PAGE_NAME(EnumBuildingTypes.MULTI_DWELLING))
```

apps/web/cypress/e2e/multi-dwelling/multi-dwelling-9.10.14-walkthrough.cy.ts
```diff
@@ -24,6 +29,19 @@ describe("multi dwelling: 9.10.14 walkthrough", () => {
...
+  it("should display all parts of breadcrumbs", () => {
+    cy.getByTestID(TESTID_BREADCRUMBS).should("be.visible");
+    cy.getByTestID(
+      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.HOME),
+    ).should("be.visible");
+    cy.getByTestID(
+      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.MULTI_DWELLING_UNIT),
+    ).should("be.visible");
+    cy.getByTestID(
+      GET_TESTID_BREADCRUMBS_BREADCRUMB(EnumBreadcrumbIds.LAST),
```
nicholaslyon-dt: I gave the last one specific unique ID since it was different. You just need:

```javascript
cy.getByTestID(TESTID_BREADCRUMB_LAST)
```

nicholaslyon-dt (commented):
@stechkev 2 notes. LMK if you want to chat about them or have thoughts.

apps/web/cypress/e2e/multi-dwelling/multi-dwelling-9.10.14-walkthrough.cy.ts
stechkev: good idea, i'll change this

packages/constants/src/constants.ts
stechkev: this is our enums for building type:
export enum EnumBuildingTypes {
  SINGLE_DWELLING = "single-dwelling",
  MULTI_DWELLING = "multi-dwelling",
}
and this is the one i made
export enum EnumBreadcrumbIds {
  HOME = "Home",
  MULTI_DWELLING_UNIT = "Multi-Dwelling Unit",
  SINGLE_DWELLING_UNIT = "Single-Dwelling Unit",
}
so i can just string builder it up, but it felt like it would be a good idea to use an enum so that when we go ahead and add more pages we could continue to add to that list. alternatively, i'd be happy if we changed the test id from data-testid="breadcrumb-Multi-Dwelling Unit" to data-testid="breadcrumb-multi-dwelling" which would let me use the existing EnumBuildingTypes without having to do any string transformations. 

The other issue is HOME, which i think i just have to hardcode somewhere as a string unless you have a better idea. if we get rid of the enum we can just have that HOME string in the test-ids like we do with breadcrumb-Last, or i can just have it hardcoded in the test, whichever is easier

stechkev (approved):

